PR #62, opened 2014-04-09
Sshkeys tag

albgus:
Added additional configuration to control the ssh key synchronization.


Phil-Friderici:
v4 does not support exporting and importing SSH keys anymore, sorry!

PR #80, opened 2014-08-08
Authorized keys command support

wolfspyre:
This adds support for an AuthorizedKeysCommand script, and the AuthorizedKeysCommandUser

you can lay the file down by specifying the template to use, or handle the file resource outside of this module.

You can have this module manage the creation of the user resource, or you can handle it elsewhere. an assumption is made about group ownership of the AuthorizedKeysCommand script.

Please let me know if you'd like any changes in order to merge this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,25 @@ Boolean to enable SendEnv options for specifying environment variables. Default 
 
 - *Default*: 'USE_DEFAULTS'
 
+sshd_config_authorized_keys_command
+----------------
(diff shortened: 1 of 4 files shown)


Phil-Friderici:
v4 does allow to set `AuthorizedKeysCommand` and `AuthorizedKeysCommandUser `in sshd_config with the `$ssh::server::authorized_keys_command` and `$ssh::server::authorized_keys_command_user` parameters.

PR #83, opened 2014-08-13
permit appending to ssh/sshd config files via params

wolfspyre:
This adds the ability to populate an array **ssh::ssh_config_appends** with data which will be appended as arbitrary lines to the `ssh_config` file.
Similarly, to append arbitrary lines to the `sshd_config` file one can populate **ssh::sshd_config_appends** with an array.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,15 @@ This will default to 'true' in future versions.
 
 - *Default*: false
 
+ssh_config_appends
+-----------------
+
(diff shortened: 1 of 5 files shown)


Phil-Friderici:
v4 does allow to add custom lines to ssh(d)_config with the help of `$ssh::custom` and `$ssh::server::custom` parameters.

PR #96, opened 2014-10-25
Debian actually disables PrintMotd by default

lazyfrosch:


Phil-Friderici:
v4 does allow to set `PrintMotd` in sshd_config with the `$ssh::server::print_motd` parameter.
The default values for different OS flavours have been set accordingly to the intention of the vendors.

PR #97, opened 2014-10-25
Use ECDSA and DSA HostKeys on Debian, as its default

lazyfrosch:
Including fallback for older versions.


Phil-Friderici:
v4 does allow to set `HostKey` in sshd_config with the `$ssh::server::host_key` parameter.
The default values for different OS flavours have been set accordingly to the intention of the vendors.

PR #102, opened 2014-11-21
WIP: Do not export an empty sshkey before sshd installation

lazyfrosch:
This avoids problems before sshd is even installed.

I actually run into this while bootstrapping an OpenSuSE box with foreman, where no sshd is installed by default.

Still missing:
- tests
- (just for me: validation if a fresh installation on OpenSUSE actually works, got still problems with the AutoYaST there)

Feel free to comment :)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -658,10 +658,12 @@ class ssh (
   }
 
   # export each node's ssh key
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Cleaning out old issues. If you are still interested in this, please re-open.

PR #105, opened 2015-04-07
Parameterize settings for SFTP usage

Phil-Friderici:
This fully backward compatible patch parameterize the settings to be used for a SFTP server setup:
ChrootDirectory
ForceCommand 
Match

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,5 @@ gem 'puppet', puppetversion
 gem 'puppetlabs_spec_helper', '>= 0.1.0'
 gem 'puppet-lint', '>= 1.0.0'
 gem 'facter', '>= 1.7.0'
+gem 'rspec-puppet', '~>1.0'
+gem 'rspec', '~>2.0'
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,6 +49,9 @@ class ssh (
   $sshd_config_allowgroups             = undef,
   $sshd_config_maxstartups             = undef,
   $sshd_config_maxsessions             = undef,
+  $sshd_config_chrootdirectory         = undef,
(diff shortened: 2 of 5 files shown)


manifests/init.pp
```diff
@@ -452,6 +455,20 @@
     }
   }
 
+  if $sshd_config_chrootdirectory != undef {
+    validate_absolute_path($sshd_config_chrootdirectory)
+  }
+
+  if $sshd_config_forcecommand != undef {
+    if is_string($sshd_config_forcecommand) == false {
```
ghoneycutt: validate_string() would have the same functionality without the extra logic

Gemfile
```diff
@@ -5,3 +5,5 @@ gem 'puppet', puppetversion
 gem 'puppetlabs_spec_helper', '>= 0.1.0'
 gem 'puppet-lint', '>= 1.0.0'
 gem 'facter', '>= 1.7.0'
+gem 'rspec-puppet', '~>1.0'
```
ghoneycutt: Is this actually needed?

Gemfile
```diff
@@ -5,3 +5,5 @@ gem 'puppet', puppetversion
...
+gem 'rspec', '~>2.0'
```
ghoneycutt: For consistency between modules, could you change the rspec line to match this

https://github.com/ghoneycutt/puppet-module-pam/blob/master/Gemfile#L14-17

ghoneycutt:
Hi @Phil-Friderici 

I went ahead and made the changes in PR #106

manifests/init.pp
Phil-Friderici: validate_string() accepts numbers as string, is_string() does not. Therefore the extra mile here.

PR #109, opened 2015-05-04
Parameterize GSSAPIAuthentication in ssh_config.

kylefu8:
Hi Garrett,

I wrote some codes to make GSSAPIAuthentication in ssh_config modifiable.
Meanwhile, I found a bug in the spec test code.
What ever I put after Puppet::Error in test code, it always passes the test.
Phil helped to fix the bug on the piece of my new tests.
But there are still defects on other lines.
I will try to fix all these lines and raise another pull request later.

Thanks!

BRs,
Kyle Fu

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -422,9 +422,7 @@ class ssh (
     fail('ssh::sshd_config_banner must be set to be able to use sshd_banner_content.')
   }
 
-  if $ssh_gssapiauthentication != undef {
-    validate_re($ssh_gssapiauthentication, '^(yes|no)$', "ssh::ssh_gssapiauthentication may be either 'yes' or 'no' and is set to <${ssh_gssapiauthentication}>.")
-  }
+  validate_re($ssh_gssapiauthentication, '^(yes|no)$', "ssh::ssh_gssapiauthentication may be either 'yes' or 'no' and is set to <${ssh_gssapiauthentication}>.")
 
   if $ssh_gssapidelegatecredentials != undef {
     validate_re($ssh_gssapidelegatecredentials, '^(yes|no)$', "ssh::ssh_gssapidelegatecredentials may be either 'yes' or 'no' and is set to <${ssh_gssapidelegatecredentials}>.")
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -24,6 +24,7 @@ class ssh (
   $ssh_config_macs                     = undef,
   $ssh_config_template                 = 'ssh/ssh_config.erb',
   $ssh_sendenv                         = 'USE_DEFAULTS',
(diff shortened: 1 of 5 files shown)


spec/classes/init_spec.rb
```diff
@@ -2254,6 +2261,39 @@
...
+        let(:params) { { :ssh_gssapiauthentication => value } }
+
+        if value.is_a?(Array)
+          value = value.join
+        end
+
+        it do
+          expect {
+            should contain_class('ssh')
+          }.to raise_error(Puppet::Error,/^ssh::ssh_gssapiauthentication may be either 'yes' or 'no' and is set to <#{value.to_s}>./)
```
ghoneycutt: Remove the period at the end. This is a matcher like similar to `*`.

spec/spec_helper.rb
```diff
@@ -1 +1,2 @@
+require 'rubygems'
```
ghoneycutt: This is not required and should not be present.

manifests/init.pp
```diff
@@ -421,6 +422,10 @@
     fail('ssh::sshd_config_banner must be set to be able to use sshd_banner_content.')
   }
 
+  if $ssh_gssapiauthentication != undef {
```
ghoneycutt: Since the variable is set to `'yes'` by default instead of `undef` then the validation must occur. Please remove the if statement and leave in the validation. Without this, you could put in `undef` and it would pass and in the template you would have GSSAPIAuthentication without a value, which would cause problems for SSH.

kylefu8: This happens to sshd_gssapiauthentication too. I think I should remove the if statement for both....

spec/spec_helper.rb
kylefu8: I found it caused by my ruby version. This line is required in Ruby 1.8...

ghoneycutt: It is not required. Check out bundler.

`gem install bundler`

and in the ssh directory type `bundle install`

and put `bundle exec` before the rspec commands, like

`bundle exec rake spec_standalone`

manifests/init.pp
ghoneycutt: Good idea. Since that would be a bugfix and your are adding a new feature, it should go into a separate branch and pull request.

spec/spec_helper.rb
kylefu8: Thanks! 
Phil helped to fix this issue by install missing gems:)

ghoneycutt:
Released in v3.27.0

kylefu8:
Thanks!
I will fix if for sshd_gssapiauthentication and spec test code in other PR.

PR #110, opened 2015-05-04
Setting mode of sshd_config file for Suse to vendor default.

ejabach:
As requested by @Phil-Friderici
Any concerns?

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -116,7 +116,7 @@ class ssh (
       $default_ssh_package_adminfile           = undef
       $default_ssh_sendenv                     = true
       $default_ssh_config_forward_x11_trusted  = 'yes'
-      $default_sshd_config_mode                = '0600'
+      $default_sshd_config_mode                = '0640'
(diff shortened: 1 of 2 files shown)


Phil-Friderici:
@ejabach thanks for taking care.
@ghoneycutt this patch will set the file mode back to the OS vendor default.
Some of Susies tools will reset the file mode back to 0640. Running into a changed state with the next Puppet run.

ghoneycutt:
I believe the default is supposed to be 0600.

Phil-Friderici:
Suse decided to change it's vendor default to 0640. If you think the module should roll out the vendor default, please merge it. If you think it should roll out the SSH close it please ;)

ghoneycutt:
I think it should default to the less permissive mode of 0600. Curious, which of Suse's tools are messing with that? You can work around this with Hiera for your Suse systems, which I'm guessing you're already doing..

Phil-Friderici:
@ghoneycutt yeah sure, we are overriding it via hiera these days.

PR #111, opened 2015-05-07
Remove if statement for sshd_gssapiauthentication.

kylefu8:
Remove if statement for sshd_gssapiauthentication to avoid undef.


ghoneycutt:
Thanks!

Published in v3.27.1

PR #113, opened 2015-05-31
Parameterize hostbasedauthentication

kTitan:
Some notes for this pull request:
* Travis is failing, because of the already existing errors in lint. Problem is that with puppetlabs_spec_helper 0.10.3 warnings will fail the lint check.
* I am not sure if the name of the variable is correct, because some of them are prefixed with "sshd_config" and some only with "sshd"
* If you prefer a squashed pull request give me a remark :)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -74,6 +74,7 @@ class ssh (
   $sshd_acceptenv                      = 'USE_DEFAULTS',
   $sshd_config_hostkey                 = 'USE_DEFAULTS',
   $sshd_listen_address                 = undef,
+  $sshd_hostbasedauthentication        = 'no',
   $service_ensure                      = 'running',
   $service_name                        = 'USE_DEFAULTS',
   $service_enable                      = true,
@@ -481,6 +482,8 @@ class ssh (
   if $sshd_config_strictmodes != undef {
     validate_re($sshd_config_strictmodes, '^(yes|no)$', "ssh::sshd_config_strictmodes may be either 'yes' or 'no' and is set to <${sshd_config_strictmodes}>.")
   }
+
+  validate_re($sshd_hostbasedauthentication, '^(yes|no)$', "ssh::sshd_hostbasedauthentication may be either 'yes' or 'no' and is set to <${sshd_hostbasedauthentication}>.")
 
   case type($hiera_merge) {
     'string': {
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -482,6 +483,10 @@
     validate_re($sshd_config_strictmodes, '^(yes|no)$', "ssh::sshd_config_strictmodes may be either 'yes' or 'no' and is set to <${sshd_config_strictmodes}>.")
   }
 
+  if $sshd_hostbasedauthentication != undef {
```
ghoneycutt: Since the default is a value and not undef, we don't need the conditional logic here.

spec/classes/init_spec.rb
```diff
@@ -2950,4 +2958,36 @@
...
+      context "specified as invalid value #{value} (as #{value.class})" do
+        let(:params) { { :sshd_hostbasedauthentication => value } }
+        if value.is_a?(Array)
+          value = value.join
+        end
+
+        it do
+          expect {
+            should contain_class('ssh')
+          }.to raise_error(Puppet::Error,/^ssh::sshd_hostbasedauthentication may be either 'yes' or 'no' and is set to <#{value}>\./)
```
ghoneycutt: remove the `^` here as I've found in my efforts to support Puppet v4 that this will cause errors as the error messaging has changed.

ghoneycutt:
Looks great! I made a few small comments. Please squash into one commit.

ghoneycutt:
I'm addressing the style issues in PR #114

ghoneycutt:
Could you rebase against master as I've merged the style fixes in PR #114

kTitan:
Ok, updated the pull request with a squashed commit, fixed the comments of you, and of course did also a rebase :)

ghoneycutt:
Apologies for all the hoops to jump through

templates/sshd_config.erb
```diff
@@ -76,7 +76,9 @@ AuthorizedKeysCommandUser <%= @sshd_authorized_keys_command_user %>
 # For this to work you will also need host keys in /etc/ssh/ssh_known_hosts
 #RhostsRSAAuthentication no
 # similar for protocol version 2
-#HostbasedAuthentication no
+<% if @sshd_hostbasedauthentication -%>
```
ghoneycutt: Missed this one. We don't need the conditional logic here for the same reason as the manifest as the value must be 'yes' or 'no' and cannot be `undef`.

ghoneycutt:
I made the small tweaks and rebased in PR #115 

Thank you for your work!!

ghoneycutt:
Released in v3.28.0

PR #118, opened 2015-08-04
fix failing spec tests

Phil-Friderici:


Phil-Friderici:
resend against support_puppet4 branch (https://github.com/ghoneycutt/puppet-module-ssh/pull/119)

PR #121, opened 2015-08-10
fix for hiera_array() always returning an empty array

Phil-Friderici:
After fixing the used variable in my last commit, I came accross that hiera_array() always delivers an empty string instead of undef on Ruby 1.8.7 :(
<pre>Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Failed to parse template ssh/sshd_config.erb:
  Filepath: /etc/puppet/environments/ssh/modules/ssh/templates/sshd_config.erb
  Line: 204
  Detail: undefined method `join' for "":String
 at /etc/puppet/environments/ssh/modules/ssh/manifests/init.pp:657 on node [redacted]
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
</pre>

With this patch the code will only by used when the variable contains an array. Adding a (silent) variable validation for free ;)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -44,10 +44,10 @@ class ssh (
   $sshd_config_banner                  = 'none',
   $sshd_config_ciphers                 = undef,
   $sshd_config_macs                    = undef,
-  $sshd_config_denyusers               = undef,
-  $sshd_config_denygroups              = undef,
-  $sshd_config_allowusers              = undef,
-  $sshd_config_allowgroups             = undef,
+  $sshd_config_allowgroups             = [],
+  $sshd_config_allowusers              = [],
+  $sshd_config_denygroups              = [],
+  $sshd_config_denyusers               = [],
   $sshd_config_maxstartups             = undef,
   $sshd_config_maxsessions             = undef,
   $sshd_config_chrootdirectory         = undef,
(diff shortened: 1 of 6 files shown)


Phil-Friderici:
btw: the error only occurs when ssh::hiera_merge is set to true.

Phil-Friderici:
/me feels like stired up a hornet's nest while fixing this. Now I have strange errors on Puppet v4 only.

Phil-Friderici:
@ghoneycutt do you have a clue what might confuses Puppet v4.0 here ?

ghoneycutt:
No clue :(

I'm OK with dropping support for 4.0.0 though. Had to do the same with 3.0.0 on some modules as the upgrade to the major version didn't go quite right and they changed things around in the .1 to fix things.

Phil-Friderici:
After spending more time I found the reason and the fix.

Reason: using hiera_array() with a default value of undef delievers changing results on different Ruby and Puppet versions.
Solution: switching to an empty array as new default value.

Now it works on all tested platforms again :)

I have added rspec tests to cover the different possible hiera_merge situations (no data, data on one level, data on multiple levels) to avoid this pithole in the future.

ghoneycutt:
@Phil-Friderici Awesome! I'm going to test locally and merge.

Phil-Friderici:
@ghoneycutt using it in our production without problems since a week

ghoneycutt:
@Phil-Friderici Thank you!

Released in v3.29.1

PR #123, opened 2015-08-12
validate_cmd support

Raskil:
added support to validate sshd_config before putting a new config file
in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -605,6 +605,20 @@ Content of root's ~/.ssh/config.
 
 - *Default*: "# This file is being maintained by Puppet.\n# DO NOT EDIT\n"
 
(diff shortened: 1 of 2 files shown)


lazyfrosch:
Wow this change is pretty heavy, a lot of indention and white space changes.

You should reduce that to an absolute minimum, and only include the actual changes. Be careful with IDEs

lazyfrosch:
@Raskil a note, you can leave that PR open, correct your changes, and force push to the same branch. This will update your PR.

Raskil:
Thanks for the note, I'll try that the next time.

ghoneycutt:
This is in PR #124

PR #126, opened 2015-09-03
Added support for Kerberos Authentication.

fang64:
I needed this for our environment, I believe I got it.

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -1002,6 +1002,7 @@ describe 'ssh' do
         :sshd_banner_content               => 'textinbanner',
         :sshd_config_xauth_location        => '/opt/ssh/bin/xauth',
         :sshd_config_subsystem_sftp        => '/opt/ssh/bin/sftp',
+        :sshd_kerberos_authentication      => 'no',
         :sshd_password_authentication      => 'no',
         :sshd_allow_tcp_forwarding         => 'no',
         :sshd_x11_forwarding               => 'no',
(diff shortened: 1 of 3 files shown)


spec/classes/init_spec.rb
```diff
@@ -1661,6 +1663,25 @@
     end
   end
 
+  context 'with sshd_kerberos_authentication set to invalid value on valid osfamily' do
```
ghoneycutt: Apologies for the spec testing code here.. it's a mess. Would you mind adding a different describe section something like this

``` ruby
describe 'with KerberosAuthentication' do
  ['yes','no'].each do |value|
    context "set to #{value}" do
      # set param to value
      it { should contain_file('sshd_config').with_content(/^KerberosAuthentication #{value}$/) }
    end
  # put this new code where you set it to invalid here
  end
end
```

ghoneycutt:
Seriously awesome work!!

Normally not a big deal, but since you'll need to rebase for the small spec test tweak, could you change the wording of the commit message? They should sound like a command, so "Add support for KerberosAuthentication option in sshd."

fang64:
@ghoneycutt Alright I'll do that.

spec/classes/init_spec.rb
fang64: so similar to this?
```puppet
describe 'with KerberosAuthentication' do
  ['yes','no','invalid'].each do |value|
    context "set to #{value}" do
      # set param to value
      it { should contain_file('sshd_config').with_content(/^KerberosAuthentication #{value}$/) }
    end
  end
end
```

I'm a little lost or don't bother with the invalid check?

ghoneycutt:
I went ahead and fixed up your commit in PR #127 

Thanks a ton for adding this feature!

PR #131, opened 2015-09-29
Add parameter sshd_addressfamily

arioch:


README.md
```diff
@@ -136,6 +136,12 @@ in ssh_config.
 
 - *Default*: undef
 
+sshd_addressfamily
+----------------
+Specifies which address family should be used by sshd
+
+- *Default*: undef
```
ghoneycutt: The default you actually use is `'any'`.

README.md
```diff
@@ -136,6 +136,12 @@ in ssh_config.
 
 - *Default*: undef
 
+sshd_addressfamily
+----------------
+Specifies which address family should be used by sshd
```
ghoneycutt: Suggest the following to show the possible values.

```
Specifies which address family should be used by sshd. Valid values are 'any', 'inet' (IPv4 only) and 'inet6' (IPv6 only).
```

manifests/init.pp
```diff
@@ -78,6 +78,7 @@
   $sshd_hostbasedauthentication        = 'no',
   $sshd_ignoreuserknownhosts           = 'no',
   $sshd_ignorerhosts                   = 'yes',
+  $sshd_addressfamily                  = 'any',
```
ghoneycutt: All params must be validated.

``` puppet
if $sshd_addressfamily != undef {
  validate_re($sshd_addressfamily, '^(any|ipv4|ipv6)$',
    "ssh::sshd_addressfamily can be undef, 'any', 'ipv4' or 'ipv6' and is set to ${sshd_addressfamily}.")
}
```

spec/classes/init_spec.rb
```diff
@@ -3241,4 +3241,16 @@
...
+  describe 'with parameter sshd_addressfamily' do
+    let(:facts) do
+      { :fqdn      => 'monkey.example.com',
+        :osfamily  => 'RedHat',
+        :sshrsakey => 'AAAAB3NzaC1yc2EAAAABIwAAAQEArGElx46pD6NNnlxVaTbp0ZJMgBKCmbTCT3RaeCk0ZUJtQ8wkcwTtqIXmmiuFsynUT0DFSd8UIodnBOPqitimmooAVAiAi30TtJVzADfPScMiUnBJKZajIBkEMkwUcqsfh630jyBvLPE/kyQcxbEeGtbu1DG3monkeymanOBW1AKc5o+cJLXcInLnbowMG7NXzujT3BRYn/9s5vtT1V9cuZJs4XLRXQ50NluxJI7sVfRPVvQI9EMbTS4AFBXUej3yfgaLSV+nPZC/lmJ2gR4t/tKvMFF9m16f8IcZKK7o0rK7v81G/tREbOT5YhcKLK+0wBfR6RsmHzwy4EddZloyLQ=='
+      }
+    end
+
+    let(:params) { { :sshd_addressfamily => 'inet' } }
+    it { should contain_file('sshd_config').with_content(/^AddressFamily inet/) }
```
ghoneycutt: add the end of line character to the regex

``` ruby
it { should contain_file('sshd_config').with_content(/^AddressFamily inet$/) }
```

spec/classes/init_spec.rb
```diff
@@ -3241,4 +3241,16 @@
       end
     end
   end
+
+  describe 'with parameter sshd_addressfamily' do
+    let(:facts) do
```
ghoneycutt: I would add something like this to test all valid values that result in the entry being in the config

``` ruby
['any','ipv4','ipv6'].each do |value|
  context "set to a valid entry of #{value}" do
     .. your code here
  end
end
```

Another context should set it to undef and show `.without_content(/^\s*AddressFamily/)`

And another context that shows if you specify an invalid value that you get an error.

This way we test that valid entries get the right config. Undef is still valid, though it ensures that the line is not present, and invalid entries cause an error.

ghoneycutt:
Hi @arioch 

Thanks for the contribution! I think you are well on the way to getting it merged. I noticed a few things were missing and made comments. Let me know if I'm not clear or you need help. Happy to do so.

arioch:
Makes sense.
I'll have a closer look tomorrow at the office.

arioch:
That should do the trick...

arioch:
I've accidentally missed a few, that should be fixed as well.
Except for one particular case: the undef one.

Could you point me in the right direction how to get it to match?
I've ended up not setting the parameter as it's undef by default. It works but it feels halfway done.

These don't seem to work:
```
let(:params) { { :sshd_addressfamily => :undef } }
let(:params) { { :sshd_addressfamily => UNDEF } }
let(:params) { { :sshd_addressfamily => nil } }
let(:params) { { :sshd_addressfamily => undef } }
let(:params) { { :sshd_addressfamily => Undef.new } }
```

ghoneycutt:
@arioch It can't be done.. on reflection, this parameter should work with all of our supported platforms, so I don't think we need the case where it is undef to remove it entirely from the config file.

manifests/init.pp
```diff
@@ -754,4 +755,9 @@
     validate_hash($keys_real)
     create_resources('ssh_authorized_key', $keys_real)
   }
+
+  if $sshd_addressfamily != undef {
+    validate_re($sshd_addressfamily, '^(any|ipv4|ipv6|inet)$',
```
ghoneycutt: `any`, `inet` and `inet6` are the supported values 

http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man5/sshd_config.5?query=sshd_config

spec/classes/init_spec.rb
```diff
@@ -3241,4 +3241,35 @@
...
+      { :fqdn      => 'monkey.example.com',
+        :osfamily  => 'RedHat',
+        :sshrsakey => 'AAAAB3NzaC1yc2EAAAABIwAAAQEArGElx46pD6NNnlxVaTbp0ZJMgBKCmbTCT3RaeCk0ZUJtQ8wkcwTtqIXmmiuFsynUT0DFSd8UIodnBOPqitimmooAVAiAi30TtJVzADfPScMiUnBJKZajIBkEMkwUcqsfh630jyBvLPE/kyQcxbEeGtbu1DG3monkeymanOBW1AKc5o+cJLXcInLnbowMG7NXzujT3BRYn/9s5vtT1V9cuZJs4XLRXQ50NluxJI7sVfRPVvQI9EMbTS4AFBXUej3yfgaLSV+nPZC/lmJ2gR4t/tKvMFF9m16f8IcZKK7o0rK7v81G/tREbOT5YhcKLK+0wBfR6RsmHzwy4EddZloyLQ=='
+      }
+    end
+
+    ['any','ipv4','ipv6','inet'].each do |value|
```
ghoneycutt: any, inet, inet6

README.md
```diff
@@ -136,6 +136,12 @@ in ssh_config.
 
 - *Default*: undef
 
+sshd_addressfamily
+----------------
+Specifies the value of the AddressFamily setting in sshd_config. Valid values are 'any', 'inet' (IPv4 only), 'inet6' (IPv6 only) and undef. A value of undef will ensure that AddressFamily is not in the configuration.
+
+- *Default*: undef
```
ghoneycutt: let's change to `any`

arioch:
That last one should do the trick...

templates/sshd_config.erb
```diff
@@ -17,7 +17,11 @@
 Port <%= @sshd_config_port %>
 #Protocol 2,1
 Protocol 2
+<% if @sshd_addressfamily != nil -%>
+AddressFamily <%= @sshd_addressfamily %>
+<% else -%>
 #AddressFamily any
```
ghoneycutt: this should always be present as we made a design choice to show the defaults, which aids in debugging a broken config

ghoneycutt:
Rebased in PR #142 

Thanks @arioch @Phil-Friderici

PR #134, opened 2015-10-15
Add a parameter to allow disabling management of the ssh service

dfairhurst:


manifests/init.pp
```diff
@@ -580,6 +581,8 @@
   }
   validate_bool($purge_keys_real)
 
+  validate_bool($manage_service)
```
ghoneycutt: Could you add the logic that allows for `'true'` as well as `true` ? See the section just above with purge_keys

spec/classes/init_spec.rb
```diff
@@ -3371,4 +3371,43 @@
...
+        let(:params) { { :manage_service => value } }
+        it do
+          expect {
+            should contain_class('ssh')
+          }.to raise_error(Puppet::Error,/is not a boolean/)
+        end
+      end
+    end
+
+    context "specified as true" do
```
ghoneycutt: test for boolean and string true. Here's an example

https://github.com/ghoneycutt/puppet-module-ntp/blob/master/spec/classes/init_spec.rb#L469

spec/classes/init_spec.rb
```diff
@@ -3371,4 +3371,43 @@
...
+      let(:params) { { :manage_service => true } }
+      it do
+        expect {
+          should contain_service('sshd_service')
+        }
+      end
+    end
+
+    context "specified as false" do
```
ghoneycutt: same for false

dfairhurst:
Updated to accept string values as requested

ghoneycutt:
Awesome work

dfairhurst:
Thanks :) and thanks for the merge.

ghoneycutt:
Released in v3.32.0

PR #137, opened 2015-11-09
Added AddressFamily option in sshd_config.

bschonec:
Currently, it's hard coded to "any".  This allows any, inet or inet6.

I've been watching and reading about how to make spec files but I'm no developer by any means.  Therefore, I've not included any changes to the spec files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,12 @@ String or Array to specify address(es) for which sshd will bind. Corresponds to 
 
 - *Default*: undef
 
+sshd_config_address_family
+-------------------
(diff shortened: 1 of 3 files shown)


Phil-Friderici:
oh, this is mostly a duplicate to https://github.com/ghoneycutt/puppet-module-ssh/pull/131

bschonec:
So is this not needed then?  I've updated the README.md but due to my utter lack of how GIT works it seems I've 'merged' other changes that I've made.

ghoneycutt:
@bschonec Thanks for the contribution though we are going with PR #131 as it already has tests and is one simple change away from being acceptable.

Phil-Friderici:
@bschonec Thanks for your work! Guess the merged change also fit your needs.

PR #144, opened 2015-12-24
Add var for MaxAuthTries

maximiliend:
This add a variable which allow to define the maximum number of authentication.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,7 +49,7 @@ class ssh (
   $sshd_config_allowusers              = [],
   $sshd_config_denygroups              = [],
   $sshd_config_denyusers               = [],
-  $sshd_config_maxauthtries            = 6,
+  $sshd_config_maxauthtries            = '6',
   $sshd_config_maxstartups             = undef,
   $sshd_config_maxsessions             = undef,
   $sshd_config_chrootdirectory         = undef,
@@ -516,6 +516,10 @@ class ssh (
 
   if $sshd_config_strictmodes != undef {
     validate_re($sshd_config_strictmodes, '^(yes|no)$', "ssh::sshd_config_strictmodes may be either 'yes' or 'no' and is set to <${sshd_config_strictmodes}>.")
+  }
+
+  if is_integer($sshd_config_maxauthtries) == false {
+    fail("ssh::sshd_config_maxauthtries must be a valid number and is set to <${sshd_config_maxauthtries}>.")
   }
 
   validate_re($sshd_hostbasedauthentication, '^(yes|no)$', "ssh::sshd_hostbasedauthentication may be either 'yes' or 'no' and is set to <${sshd_hostbasedauthentication}>.")
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,6 +49,7 @@ class ssh (
   $sshd_config_allowusers              = [],
   $sshd_config_denygroups              = [],
   $sshd_config_denyusers               = [],
+  $sshd_config_maxauthtries            = 6,
   $sshd_config_maxstartups             = undef,
   $sshd_config_maxsessions             = undef,
   $sshd_config_chrootdirectory         = undef,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -48,6 +48,7 @@
   $sshd_config_allowusers              = [],
   $sshd_config_denygroups              = [],
   $sshd_config_denyusers               = [],
+  $sshd_config_maxauthtries            = 5,
```
ghoneycutt: It appears the default was 6, so that should remain

templates/sshd_config.erb
```diff
@@ -61,7 +61,10 @@ PermitRootLogin <%= @permit_root_login %>
 <% if @sshd_config_strictmodes -%>
 StrictModes <%= @sshd_config_strictmodes %>
 <% end -%>
-#MaxAuthTries 6
```
ghoneycutt: We keep the default commented out to aid in debugging

ghoneycutt:
Thanks @maximiliend 

Could you please document the new parameter in the README? We also need spec tests in order to accept. Look at a similar param and copy how it is tested.

maximiliend:
Thank you for your advice, the defaults values are amended and the README enhanced. I'll look for specific tests.

spec/classes/init_spec.rb
```diff
@@ -608,6 +608,7 @@
     it { should contain_file('sshd_config').with_content(/^AcceptEnv L.*$/) }
     it { should_not contain_file('sshd_config').with_content(/^AuthorizedKeysFile/) }
     it { should_not contain_file('sshd_config').with_content(/^StrictModes/) }
+    it { should contain_file('sshd_config').with_content(/^MaxAuthTries/) }
```
ghoneycutt: This should have a `6` in it, as by default the line should have that

spec/classes/init_spec.rb
```diff
@@ -2045,6 +2047,25 @@
     end
   end
 
+  context 'with sshd_config_maxauthtries not being a valid number' do
```
ghoneycutt: To get this to work, we need to validate the parameter in the ssh class. Something like

`validate_integer()`

manifests/init.pp
```diff
@@ -48,6 +48,7 @@
   $sshd_config_allowusers              = [],
   $sshd_config_denygroups              = [],
   $sshd_config_denyusers               = [],
+  $sshd_config_maxauthtries            = 6,
```
ghoneycutt: Sometimes people will inadvertently use a string `'6'`. To deal with this we can do something like

``` puppet
$sshd_config_maxauthtries_real = $sshd_config_maxauthtries + 0
validate_integer($sshd_config_maxauthtries_real)
```

spec/classes/init_spec.rb
maximiliend: I understood ;)

spec/classes/init_spec.rb
maximiliend: Ok or maybe something like :
validate_re($sshd_config_maxauthtries, '^\d+$', "ssh::sshd_config_maxauthtries must be a valid number and is set to <${sshd_config_maxauthtries}>.")
Thanks you.

manifests/init.pp
maximiliend: Got it too ! I'll use directly '6'. But thanks for the advice.

Thanks a lot for your help. Sorry for the mistakes that I made. Tell me if there is something else.

I'll push again my commit.

spec/classes/init_spec.rb
ghoneycutt: validate_re() expects a string and it really should be an integer

spec/classes/init_spec.rb
```diff
@@ -608,6 +608,7 @@
     it { should contain_file('sshd_config').with_content(/^AcceptEnv L.*$/) }
     it { should_not contain_file('sshd_config').with_content(/^AuthorizedKeysFile/) }
     it { should_not contain_file('sshd_config').with_content(/^StrictModes/) }
+    it { should contain_file('sshd_config').with_content(/^MaxAuthTries 6/) }
```
ghoneycutt: There are a handful of other spots that a line similar to this one should be in.

maximiliend:
I don't know if I can add more spec tests. I hope this is enough.

maximiliend:
Is there anything else missing ?

jthiesfeld:
What is the status of this patch as it would be useful in my environment as well?

ghoneycutt:
@jthiesfeld the spec tests were refactored and so this cannot be merged due to that. You want to take a crack at it?

jthiesfeld:
I'm new to puppet, ruby, and git.

It looks like maximiliend has a few pull requests, but just the rspec
change didn't get accepted?  Any pointers on how to get started?

PR #147, opened 2016-01-14
mitigate client bugs CVE-2016-0777 and CVE-2016-0778

mjm1138:
Sets the undocumented "UseRoaming" parameter to "no" in sshd_config

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -22,6 +22,7 @@ class ssh (
   $ssh_config_sendenv_xmodifiers       = false,
   $ssh_config_ciphers                  = undef,
   $ssh_config_macs                     = undef,
+  $ssh_config_use_roaming              = 'no',
   $ssh_config_template                 = 'ssh/ssh_config.erb',
   $ssh_sendenv                         = 'USE_DEFAULTS',
   $ssh_gssapiauthentication            = 'yes',
--- a/templates/ssh_config.erb
+++ b/templates/ssh_config.erb
@@ -72,6 +72,9 @@ GSSAPIDelegateCredentials <%= @ssh_gssapidelegatecredentials %>
 <% if @ssh_config_forward_x11 != nil -%>
   ForwardX11 <%= @ssh_config_forward_x11 %>
 <% end -%>
+<% if @ssh_config_use_roaming != nil -%>
+  UseRoaming <%= @ssh_config_use_roaming %>
+<% end -%>
 <% if @ssh_config_server_alive_interval != nil -%>
   ServerAliveInterval <%= @ssh_config_server_alive_interval %>
 <% end -%>


ghoneycutt:
Thanks @mjm1138 

Could you please provide some documentation around this undocumented feature? I've seen when introducing a config option that is not known to sshd that instead of the parser skipping it, the daemon will refuse to start. I want to ensure that would not happen here as we support many platforms.

ghoneycutt:
@anders-larsson @Phil-Friderici would you mind testing your matrix of systems before we merge this. I want to ensure we don't break anything for you.

anders-larsson:
Hi,

I ran it on all osfamily + operatingsystemrelease combinations in our environment and everything seemed fine even on openssh versions in which UseRoaming system is not implemented.

Applied this change, restarted ssh and tried to connect to sshd on the following osfamily+operatingsystem combinations. No errors could be find in logs either.
``````
Debian-12.04
Debian-14.04
RedHat-5.11 
RedHat-5.9
RedHat-6.1
RedHat-6.2
RedHat-6.3
RedHat-6.4
RedHat-6.5
RedHat-6.6
RedHat-6.7
RedHat-7.0
RedHat-7.1
RedHat-7.2
Suse-10.2
Suse-10.3
Suse-10.4
Suse-11.0
Suse-11.1
Suse-11.2
Suse-11.3
Suse-11.4
Suse-12.0
``````

Phil-Friderici:
I have added the details that I feel were missing in https://github.com/ghoneycutt/puppet-module-ssh/pull/148

Phil-Friderici:
@nalyanyam Could you please do a test run if Solaris would have an issue when the client configuration ssh_config does contain this additional line:

<pre>
UseRoaming no
</pre>

nalyanyam:
@Phil-Friderici 

This option does not work on Solaris. The sshd daemon is not affected as expected, but ssh client fails.

grep -i UseRoaming /etc/ssh/ssh_config
UseRoaming no
Host [13:37] [/home/xhpnama] -> ssh hostname
/etc/ssh/ssh_config: line 56: unknown configuration option: UseRoaming
terminating, 1 bad configuration option(s)

nalyanyam:
If this option is supported by other OS, then it should be configured as optional - with the "UseRoaming no/yes" configuration option absent for  Solaris by default.

Phil-Friderici:
@nalyanyam THANKS A LOT for helping us again. I'll update my PR https://github.com/ghoneycutt/puppet-module-ssh/pull/148 to not set this parameter on Solaris systems.

anders-larsson:
@Phil-Friderici oh.. My tests are actually wrong. We need to decide whether to put it in the configuration file or not based on the version of openssh. It does not work on versions that it isn't implemented in. I guess we need to add a fact with the openssh version.

Phil-Friderici:
@anders-larsson hmmmm too bad, can't continue to work on that before end of next week :(
Can you take over ? Recommending to have a look at my work in PR https://github.com/ghoneycutt/puppet-module-ssh/pull/148

mjm1138:
@ghoneycutt OpenSSH release notes mention the feature here: http://www.openssh.com/txt/release-7.1p2

So notably this flag is not available on versions < 5.4, and presumably not on non-OpenSSH implementations.

ghoneycutt:
@mjm1138 awesome work! We have people using older versions and non-OpenSSH implementations. Agree with @anders-larsson that we need a custom fact. Going to add a bit to the README in Phil's PR #148 and merge

ghoneycutt:
Thank you everyone!

Closing in favor of #149 which contains this commit

PR #151, opened 2016-01-22
Improved control over exported host keys

apatik:
Added a new ssh_key_export parameter to control whether the node exports its SSH host key (mirroring the existing ssh_key_import parameter). Defaults to true for backwards compatibility.

Added a new ssh_key_host_aliases parameter to allow passing host aliases to the exported sshkey declaration (allows setting e.g. hostname and IP address as aliases alongside the FQDN). Defaults to undef for backwards compatibility.

Expanded ssh_key_type parameter checks to support ed25519 and ecdsa keys - all listed type strings are pulled straight from the Puppet Type Reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,6 +586,18 @@ Export node SSH key. Valid values are 'present' and 'absent'.
 
 - *Default*: 'present'
 
+ssh_key_export
+--------------
+Controls whether node SSH host key should be exported. Valid values are 'true and 'false'.
+
+- *Default*: 'true'
+
+ssh_key_host_aliases
+--------------
+Specifies any alias(es) for the node's exported host key. Passed directly to the sshkey exported resource. Valid values are a string or array of strings.
+
+- *Default*: undef
+
 ssh_key_import
 --------------
 Import all exported node SSH keys. Valid values are 'true' and 'false'.
@@ -594,7 +606,7 @@ Import all exported node SSH keys. Valid values are 'true' and 'false'.
 
 ssh_key_type
 ------------
-Encryption type for SSH key. Valid values are 'rsa', 'dsa', 'ssh-dss' and 'ssh-rsa'
+Encryption type for SSH key. Valid values are 'rsa', 'dsa', 'ssh-dss', 'ssh-rsa', 'ssh-ed25519', 'ed25519', 'ecdsa-sha2-nistp256', 'ecdsa-sha2-nistp384', and 'ecdsa-sha2-nistp521'
 
 - *Default*: 'ssh-rsa'
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -595,7 +595,7 @@ class ssh (
       $key = $::sshdsakey
     }
     'ssh-ed25519', 'ed25519': {
-      $key = $::sshed25519key 
+      $key = $::sshed25519key
     }
     'ecdsa-sha2-nistp256', 'ecdsa-sha2-nistp384', 'ecdsa-sha2-nistp521': {
       $key = $::sshecdsakey
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -87,6 +87,8 @@ class ssh (
   $service_hasrestart                  = true,
   $service_hasstatus                   = 'USE_DEFAULTS',
   $ssh_key_ensure                      = 'present',
+  $ssh_key_export                      = true,
+  $ssh_key_host_aliases                = undef,
   $ssh_key_import                      = true,
   $ssh_key_type                        = 'ssh-rsa',
   $ssh_config_global_known_hosts_file  = '/etc/ssh/ssh_known_hosts',
@@ -536,6 +538,20 @@ class ssh (
     }
   }
 
+  case type3x($ssh_key_export) {
+    'string': {
+      validate_re($ssh_key_export, '^(true|false)$', "ssh::ssh_key_export may be either 'true' or 'false' and is set to <${ssh_key_export}>.")
+      $ssh_key_export_real = str2bool($ssh_key_export)
+    }
+    'boolean': {
+      $ssh_key_export_real = $ssh_key_export
+    }
+    default: {
+      fail('ssh::ssh_key_export type must be true or false.')
+    }
+  }
+  validate_bool($ssh_key_export_real)
+
   case type3x($ssh_key_import) {
     'string': {
       validate_re($ssh_key_import, '^(true|false)$', "ssh::ssh_key_import may be either 'true' or 'false' and is set to <${ssh_key_import}>.")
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -746,10 +768,13 @@
   }
 
   # export each node's ssh key
-  @@sshkey { $::fqdn :
```
ghoneycutt: I believe if you don't have exported resources enabled with PuppetDB, this is just silently skipped. Great having the ability to not export it though, especially if using PuppetDB, to keep unwanted bits out of your database.

spec/classes/init_spec.rb
```diff
@@ -3285,6 +3285,63 @@
...
   end
+  
+  describe 'with ssh_key_export parameter specified' do
+    context 'as a non-boolean or non-string' do
+    let(:params) { { :ssh_key_export => ['not_a_boolean','or_a_string'] } }
+
+      it 'should fail' do
+        expect {
+          should contain_class('ssh')
+        }.to raise_error(Puppet::Error)
```
ghoneycutt: please put the actual error message in here. A way to find out is to put 

```ruby
}.to raise_error(Puppet::Error,/asdf/)
```

and learn from the spec failure

spec/classes/init_spec.rb
```diff
@@ -3285,6 +3285,63 @@
...
+    ['true',true].each do |value|
+      context "as #{value}" do
+        let(:params) { { :ssh_key_export => value } }
+        let(:facts) do
+          { :osfamily          => 'RedHat',
+            :lsbmajdistrelease => '6',
+          }
+        end
+
+        it { should compile.with_all_deps }
```
ghoneycutt: Since ssh_key_import defaults to true, the sshkey that you export should be in the catalog. Could you try adding a resource for the sshkey for here with all the values, including the new host_aliases?

spec/classes/init_spec.rb
```diff
@@ -3285,6 +3285,63 @@
...
+
+    ['false',false].each do |value|
+      context "as #{value}" do
+        let(:params) { { :ssh_key_export => value } }
+        let(:facts) do
+          { :osfamily          => 'RedHat',
+            :lsbmajdistrelease => '6',
+          }
+        end
+
```
ghoneycutt: similar to above, just with the resource not existing

ghoneycutt:
PR with documentation **and** spec tests :) This is awesome!!

spec/classes/init_spec.rb
apatik: When I put
it { $stdout.puts self.catalogue.to_yaml }
at this location to view the catalogs that rspec is using, the exported sshkey isn't in there, and any test I try to write against it (obviously) fails. Do you see something different if you try it? I'm not sure we can test for the existence of resources in an export / import paradigm this way (although I would love to be proven wrong on that).

ghoneycutt: Doh.. exported resources are still not supported by rspec-puppet, so you're right that they would not show up.

manifests/init.pp
```diff
@@ -536,6 +538,20 @@
     }
   }
 
+  case type3x($ssh_key_export) {
+    'string': {
+      validate_re($ssh_key_export, '^(true|false)$', "ssh::ssh_key_export may be either 'true' or 'false' and is set to <${ssh_key_export}>.")
```
ghoneycutt: This is unnecessary since the validate_bool() below will catch it

manifests/init.pp
```diff
@@ -536,6 +538,20 @@
...
+      $ssh_key_export_real = str2bool($ssh_key_export)
+    }
+    'boolean': {
+      $ssh_key_export_real = $ssh_key_export
+    }
+    default: {
+      fail('ssh::ssh_key_export type must be true or false.')
```
ghoneycutt: This will muddy the error being thrown and make spec testing harder as there would be two different error strings

manifests/init.pp
```diff
@@ -536,6 +538,20 @@
     }
   }
 
+  case type3x($ssh_key_export) {
```
ghoneycutt: Suggest this pattern

```puppet
if is_string($ssh_key_export) == true {
  $ssh_key_export_real = str2bool($ssh_key_export)
} else {
  $ssh_key_export_real = $ssh_key_export
}
validate_bool($ssh_key_export_real)
```

README.md
```diff
@@ -586,6 +586,18 @@ Export node SSH key. Valid values are 'present' and 'absent'.
 
 - *Default*: 'present'
 
+ssh_key_export
+--------------
+Controls whether node SSH host key should be exported. Valid values are 'true and 'false'.
```
ghoneycutt: Could you please change to `Boolean to determine if the node's SSH host key should be exported.`

I think that's a bit clearer and encourages booleans as opposed to stringified booleans.

README.md
```diff
@@ -594,7 +606,7 @@ Import all exported node SSH keys. Valid values are 'true' and 'false'.
 
 ssh_key_type
 ------------
-Encryption type for SSH key. Valid values are 'rsa', 'dsa', 'ssh-dss' and 'ssh-rsa'
+Encryption type for SSH key. Valid values are 'rsa', 'dsa', 'ssh-dss', 'ssh-rsa', 'ssh-ed25519', 'ed25519', 'ecdsa-sha2-nistp256', 'ecdsa-sha2-nistp384', and 'ecdsa-sha2-nistp521'
```
ghoneycutt: To future proof this

``` markdown
Encryption type for SSH key. Valid values are 'rsa', 'dsa', 'ssh-dss', 'ssh-rsa', 'ssh-ed25519',
'ed25519', 'ecdsa-sha2-nistp256', 'ecdsa-sha2-nistp384', 'ecdsa-sha2-nistp521' and anything
 else that is valid for the `type` attribute of the [sshkey](http://docs.puppetlabs.com/references/stable/type.html#sshkey) resource type.
```

apatik: Except that's not entirely true - the ssh_key_type is matched against the currently listed values to determine which host key fact to use - if a new type is added by Puppet labs, using it will cause a compilation failure unless the module is updated to explicitly support it.

manifests/init.pp
apatik: I literally copy + pasted the ssh_key_import code and changed the variable names - if you want it written better, that's not a big deal, I'll do it. Just realize there's more of this in there.

manifests/init.pp
apatik: While this will be shorter and will likely make the spec tests a bit simpler, if someone using the module makes a typo or otherwise enters an invalid value, it's going to throw either "Puppet::Error: str2bool(): Unknown type of boolean given" or "Puppet::Error: 'theTypodString' is not a boolean.", which doesn't really explain the issue to the user - it just looks like the module is broken internally. I would argue in favor of having the tests be slightly more complex in order for the error messages to be more meaningful to a new user of the module who might not understand the internals of it (and frankly, shouldn't have to to debug a compilation error).

README.md
ghoneycutt: yeah, we'll need to change that ..

sp-joseluis-ledesma:
What is missing to be able to merge this PR?

PR #154, opened 2016-02-15
Specs for ssh* facts

Phil-Friderici:
The spec tests for ssh* facts got lost while merging.
I have added support for semantic version strings as well.


Phil-Friderici:
@ghoneycutt I recommend to merge this to have tests for the added ssh_version* facts.

PR #157, opened 2016-02-24
UseRoaming on supported versions only

anders-larsson:
Hi,

This pull request makes sure that UseRoaming is only configured on Openssh version where it's supported. Since the new facts ssh_version and ssh_version_numeric is used it required a few changes in the spec tests (add facts to all tests) had to be rewritten a biit to make it easier in the further to add new spec tests.

I hope it looks OK.


ghoneycutt:
@anders-larsson Wow! What a refactor, looks good though I have not dived in too deep. Have you tested this internally?

@Phil-Friderici You good with this?

anders-larsson:
@ghoneycutt The UseRoaming part has been tested on all combinations of osfamily+operatingsystem-release (Linux only though). Would be good if someone could test it on Solaris since we're quite not there yet on the Solaris part in managing stuff with Puppet.

Looked good. Regarding the spec tests they seem to work OK. Would be awesome if someone else can verify the difference between them since it's quite a big change. Might be worth to verify all checks are supposed to fail are actually failing with the right reason and not but because there are facts missing.

Phil-Friderici:
@ghoneycutt Yes I like the new functionality it adds to UseRoadming logic and makes good use of the new ssh_version* facts. However, I wasn't able to get into the details of the tests myself.

ghoneycutt:
@Phil-Friderici would you have time soon to check them out? I know you will see everything related to the spec tests that I may not.

Phil-Friderici:
@ghoneycutt: I would love to... it's just the schedules are tough at the moment.
On the other side this would be a nice use case to give the Github fat client a try...

ghoneycutt:
@anders-larsson If you want a quick merge of the useroaming functionality, suggest sending a PR without the big refactor and doing the refactor in a separate PR. The useroaming should be simple to merge while the big refactor will take more time.

Phil-Friderici:
I did start with comparing the spec.

Verified successfully so far:
- ssh with default params on
  - RH5/6/7
  - Solaris 5.9/5.10/.511

Phil-Friderici:
Verified successfull so far:
- ssh with default params on
  - RH5/6/7
  - Solaris 5.9/5.10/5.11
  - Debian

While verifing Suse I realized that ```UseRoaming``` isn't set in ssh_config which was the case beforehand. Is it turned off intenionally ?

Phil-Friderici:
Verified successfull so far:
- ssh with default params on
  - RH5/6/7
  - Solaris 5.9/5.10/5.11
  - Debian
  - Suse i386 / x86_64 (doesn't use UseRoadming at all, please verify)

anders-larsson:
Looks OK for me on SLE:

`````
release:ssh_version:value
10.2:4.2:unset
10.3:4.2:unset
10.4:5.1:unset
11.0:5.1:unset
11.1:5.1:unset
11.2:5.1:unset
11.3:6.2:no
11.4:6.6:no
12.0:6.6:no
12.1:6.6no
`````

Phil-Friderici:
I found ```UseRoaming``` only in the fixtures for Debian and RH:
<pre>
# pwd; grep UseRoaming *
./puppet-module-ssh/spec/fixtures
ssh_config_debian:  UseRoaming no
ssh_config_rhel:  UseRoaming no
</pre>

anders-larsson:
Hmm. Sounds like there's something wrong with the spec tests in that case. Because the real life test I made seems to be correct. All releases over 5.4 should have UseRoaming set to 'no'.

I'll look into it!

anders-larsson:
Hi,

Found the bug. I had accidentally missed to prepend OpenSSH_ to the fact ssh_version on some of the Suse systems. I've updated and fixed it now. Good find @Phil-Friderici.

Phil-Friderici:
to fix errors on Ruby 1.8.7 have a look at this:
https://github.com/kylefu8/puppet-module-tweaks/commit/7d4cfab307987c44879ee94964b7ef12fcdaa803

Phil-Friderici:
@ghoneycutt the refactoring does introduce two awesome changes:
- loops through OSfamilies to follow DRY on large scale :clap:
- uses default facts with merge for specifics deviations

After UseRoaming testing got fixed, it get's my  :ok_hand: :+1:

Phil-Friderici:
@anders-larsson: I didn't even know about a 140 char check
@ghoneycutt: looks good to me

Phil-Friderici:
@ghoneycutt This is should be no push, but I would be happy if you could merging this patch before others. A rewrite & re-review would be quite a lot of work.

Thanks & have phun
Phil

ghoneycutt:
Outstanding! Modernizing the spec tests are very very much appreciated!

ghoneycutt:
Released in v3.36.0

Phil-Friderici:
Thanks for your work gentlemen !

Issue #130, opened 2015-09-24
sshd_config allows multiple instances in config, but sshd_config_port is single valued 


sbroddy:
Hello, 

In our environment, we run sshd on multiple ports.  The sshd config allows for this (from the manpage):

      Port    Specifies the port number that sshd(8) listens on.  
      The default is 22.  Multiple options of this type are permitted.  
      See also ListenAddress.

How hard would it be to modify things so that sshd_config_port would accept an array?  Such as (in hiera):

ssh::sshd_config_port:
  - '22'
  - '1111'

I'm just getting started with hiera, so there may be another easier way to accomplish what I need, but at present it looks to me like this module will not accept multiple values for this configuration option.

Thank you for providing some great modules!

bschonec:
I'm thinking the same thing.  I can modify the code but I am *not* qualified to write spec tests.   I have no idea how to do that.

ghoneycutt:
I can help with the spec tests. Can you please send a PR for accepting an array of ports? Be sure to also update the README when doing so.

ghoneycutt:
Functionality added in v3.40.0

PR #158, opened 2016-03-10
adding sshd_config_ports param to permit multiple listen ports

westbywest:
This PR, inspired by https://github.com/ghoneycutt/puppet-module-ssh/issues/130 , adds a new param **sshd_config_ports** to permit specification of additional ports in sshd_config.  This approach, while less than elegant, allows backward compatibility for those using the existing param **sshd_config_port** .

Also tweaked validation on **sshd_config_port**, as it would fail when passed a Fixnum value, i.e. when supplied via hiera.

P.S. I chose not to try to implement **sshd_config_port** as a polymorphic parameter (i.e. would accept either Fixnum, String, or Array), since that would erode the module's ability to validate its parameter inputs. Indeed, despite the README stating such, the parameter **sshd_listen_address** actually isn't polymorphic; [will only accept an array](https://github.com/ghoneycutt/puppet-module-ssh/blob/master/manifests/init.pp#L394).

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,12 @@ String to specify listen port for sshd. Port option in sshd_config.
 
 - *Default*: '22'
 
+sshd_config_ports
+---------------------------
+Array to specify optional, additional listen ports for sshd.  Should not include value specified above in sshd_config_port.
+
+- *Default*: undef
+
 sshd_config_syslog_facility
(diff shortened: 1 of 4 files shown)


ghoneycutt:
Functionality added in v3.40.0

PR #162, opened 2016-05-10
Change sshd_config_port to support multiple port numbers

andrei693:
Changed validate_re to validate_array and modified the template accordingly.

Fixes GH-130 in a different way than [PR-158](https://github.com/ghoneycutt/puppet-module-ssh/pull/158)

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ String or Array to specify address(es) for which sshd will bind. Corresponds to 
 
 sshd_config_port
 ---------------------------
-String to specify listen port for sshd. Port option in sshd_config.
+String, Integer or Array to specify listen port[s] for sshd. Port option in sshd_config.
 
 - *Default*: '22'
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -442,7 +442,22 @@ class ssh (
   if $ssh_config_hash_known_hosts_real != undef {
     validate_re($ssh_config_hash_known_hosts_real, '^(yes|no)$', "ssh::ssh_config_hash_known_hosts may be either 'yes' or 'no' and is set to <${ssh_config_hash_known_hosts_real}>.")
   }
-  validate_re($sshd_config_port, '^\d+$', "ssh::sshd_config_port must be a valid number and is set to <${sshd_config_port}>.")
+  case type3x($sshd_config_port) {
+    'string': {
+      validate_re($sshd_config_port, '^\d+$', "ssh::sshd_config_port must be a valid number and is set to <${sshd_config_port}>.")
+      $sshd_config_port_array = [ str2num($sshd_config_port) ]
+    }
+    'array': {
+      $sshd_config_port_array = $sshd_config_port
+    }
+    'integer': {
+      $sshd_config_port_array = [ $sshd_config_port ]
+    }
+    default: {
+      fail('ssh:sshd_config_port must be a string, an integer or an array. ')
+    }
+  }
+  validate_numeric($sshd_config_port_array, 65535, 1)
   if $sshd_kerberos_authentication != undef {
     validate_re($sshd_kerberos_authentication, '^(yes|no)$', "ssh::sshd_kerberos_authentication may be either 'yes' or 'no' and is set to <${sshd_kerberos_authentication}>.")
   }
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -950,21 +950,40 @@ describe 'ssh' do
     end
   end
 
-  context 'with sshd_config_port not being a valid number' do
+  describe 'sshd_config_port param' do
     let :facts do
       default_facts.merge(
         {
         }
       )
     end
-    let :params do
-      { :sshd_config_port => '22invalid' }
+
+    context 'when set to an array' do
+      let (:params) {{'sshd_config_port' => ['22222', '22223'] }}
+
+      it { should contain_file('sshd_config').with_content(/^Port 22222\nPort 22223$/) }
     end
 
-    it 'should fail' do
-      expect {
-        should contain_class('ssh')
-      }.to raise_error(Puppet::Error,/ssh::sshd_config_port must be a valid number and is set to <22invalid>\./)
+    context 'when set to a string' do
+      let (:params) {{'sshd_config_port' => '22222' }}
+
+      it { should contain_file('sshd_config').with_content(/^Port 22222$/) }
+    end
+
+    context 'when set to an integer' do
+      let (:params) {{'sshd_config_port' => 22222 }}
+
+      it { should contain_file('sshd_config').with_content(/^Port 22222$/) }
+    end
+
+    context 'when not set to a valid number' do
+      let (:params) {{'sshd_config_port' => '22invalid' }}
+
+      it 'should fail' do
+        expect {
+          should contain_class('ssh')
+        }.to raise_error(Puppet::Error,/ssh::sshd_config_port must be a valid number and is set to <22invalid>\./)
+      end
     end
   end
 
--- a/templates/sshd_config.erb
+++ b/templates/sshd_config.erb
@@ -14,7 +14,9 @@
 # default value.
 
 #Port 22
-Port <%= @sshd_config_port %>
+<% @sshd_config_port_array.each do |p| -%>
+<%= "Port #{p}" %>
+<% end -%>
 #Protocol 2,1
 Protocol 2
 #AddressFamily any


README.md
```diff
@@ -196,9 +196,9 @@ String or Array to specify address(es) for which sshd will bind. Corresponds to
 
 sshd_config_port
 ---------------------------
-String to specify listen port for sshd. Port option in sshd_config.
+Array to specify listen port[s] for sshd. Port option in sshd_config.
 
-- *Default*: '22'
+- *Default*: ['22']
```
ghoneycutt: This would need a major version change as it changes the API in semver speak. To maintain backward compatibility sshd_config_port would need to support being a string or an array.

templates/sshd_config.erb
```diff
@@ -12,9 +12,13 @@
...
 # possible, but leave them commented.  Uncommented options change a
 # default value.
-
+<% if port = scope.lookupvar("ssh::sshd_config_port") and port.any? %>
+<% port.each do |p| -%>
+<%= "Port #{p}" %>
+<% end -%>
+<%else -%>
 #Port 22
-Port <%= @sshd_config_port %>
```
ghoneycutt: You could use `.class` and if it is string do variable interpolation and if an array then iteration.

ghoneycutt: Check out https://github.com/ghoneycutt/puppet-module-ssh/blob/master/templates/sshd_config.erb#L24

README.md
```diff
@@ -196,9 +196,9 @@ String or Array to specify address(es) for which sshd will bind. Corresponds to
 
 sshd_config_port
 ---------------------------
-String to specify listen port for sshd. Port option in sshd_config.
+Array to specify listen port[s] for sshd. Port option in sshd_config.
```
ghoneycutt: For validation you could do something like

```puppet
if is_type($sshd_config_port) != 'String' and is_type($sshd_config_port) != 'Array' {
  fail('sshd_config_port must be an array or a string')
}
```

andrei693:
Sorry for the late reply. Changed to accept both string and array now. Let me know if there is anything else.

Thx

spec/classes/init_spec.rb
```diff
@@ -950,22 +950,29 @@
...
-      }.to raise_error(Puppet::Error,/ssh::sshd_config_port must be a valid number and is set to <22invalid>\./)
-    end
+  context 'when set to an array' do
+    let (:params) {{'sshd_config_port' => ['22222', '22223'] }}
+
+    it { should contain_file('sshd_config').with_content(/^Port 22222\nPort 22223$/) }
+  end
+
+  context 'when set to a string' do
+    let (:params) {{'sshd_config_port' => ['22222'] }}
```
ghoneycutt: This is an array, not a string, for which the test is meant.

andrei693: This was my first attempt at rspec. I thought it will test sshd_config_port as an array and as a string. From reading it again I think it tests if it's an array with 2 values and then if it's an array with 1 value. I copied the format from https://github.com/ghoneycutt/puppet-module-ssh/blob/master/spec/classes/init_spec.rb#L599.

spec/classes/init_spec.rb
```diff
@@ -950,22 +950,29 @@
...
-    let :params do
-      { :sshd_config_port => '22invalid' }
-    end
 
-    it 'should fail' do
-      expect {
-        should contain_class('ssh')
-      }.to raise_error(Puppet::Error,/ssh::sshd_config_port must be a valid number and is set to <22invalid>\./)
-    end
+  context 'when set to an array' do
```
ghoneycutt: `when set to an array with multiple values`

spec/classes/init_spec.rb
```diff
@@ -950,22 +950,29 @@
...
+    let (:params) {{'sshd_config_port' => ['22222', '22223'] }}
+
+    it { should contain_file('sshd_config').with_content(/^Port 22222\nPort 22223$/) }
+  end
+
+  context 'when set to a string' do
```
ghoneycutt: `when set to an array with a single value`

spec/classes/init_spec.rb
ghoneycutt: Those two tests are good, though they need renamed a bit.

Needs an additional test where sshd_config_port is a string. A string looks like this. Notice no square brackets.

``` ruby
let (:params) {{'sshd_config_port' => '22222' }}
```

andrei693:
Reworded the array tests and added a string test.

andrei693:
Anything I can do to have this moving forward?

spec/classes/init_spec.rb
```diff
@@ -950,22 +950,34 @@
...
+
+    it { should contain_file('sshd_config').with_content(/^Port 22222\n/) }
+  end
+
+  context 'when set to a string' do
+    let (:params) {{'sshd_config_port' => '22222' }}
+
+    it { should contain_file('sshd_config').with_content(/^Port 22222\n/) }
+
+  context 'when not set' do
```
ghoneycutt: It will always be set as it is validated such that it is required

ghoneycutt:
Hi @andrei693 

Please remove the test for port not being specified. It seems there is a syntax error with your tests. You can validate it with `ruby -c`. Thanks again for working on adding this functionality.

andrei693:
Ugh, at last I managed to get it working after abusing Travis a bit. Please have another look and see if it's worth.

Thx

ghoneycutt:
@anders-larsson @Phil-Friderici this look OK to you? I'd like to get some more eyes on it before I merge.

manifests/init.pp
```diff
@@ -442,7 +442,20 @@
   if $ssh_config_hash_known_hosts_real != undef {
     validate_re($ssh_config_hash_known_hosts_real, '^(yes|no)$', "ssh::ssh_config_hash_known_hosts may be either 'yes' or 'no' and is set to <${ssh_config_hash_known_hosts_real}>.")
   }
-  validate_re($sshd_config_port, '^\d+$', "ssh::sshd_config_port must be a valid number and is set to <${sshd_config_port}>.")
+  case type3x($sshd_config_port) {
+    'string': {
+      validate_re($sshd_config_port, '^\d+$', "ssh::sshd_config_port must be a valid number and is set to <${sshd_config_port}>.")
+    }
+    'array': {
+      validate_array($sshd_config_port)
```
anders-larsson: Correct me if I'm wrong but do you really need to validate that sshd_config_port is an array if the type3x functions returns that it is an array?

I would propose that we also use validate_numeric on the array to make sure the array only contain values with numbers between 1 to 65525. Looking at stdlibs documentation it should work. https://github.com/puppetlabs/puppetlabs-stdlib#validate_numeric

anders-larsson:
Functionality wise it works great!

README.md
```diff
@@ -196,7 +196,7 @@ String or Array to specify address(es) for which sshd will bind. Corresponds to
 
 sshd_config_port
 ---------------------------
-String to specify listen port for sshd. Port option in sshd_config.
+String or Array to specify listen port[s] for sshd. Port option in sshd_config.
```
Phil-Friderici: I like to see the support for integers here :+1: 
Please add the good news here too :)

templates/sshd_config.erb
```diff
@@ -14,7 +14,13 @@
 # default value.
 
 #Port 22
+<% if @sshd_config_port.class == Array -%>
+<% @sshd_config_port.each do |p| -%>
+<%= "Port #{p}" %>
+<% end -%>
+<% else -%>
 Port <%= @sshd_config_port %>
+<% end -%>
```
Phil-Friderici: Just a side comment:
usually I try to avoid functional logic in templates. In this case you could normalize the validated data to become an array before it reaches the template.

spec/classes/init_spec.rb
```diff
@@ -950,21 +950,34 @@
...
-      expect {
-        should contain_class('ssh')
-      }.to raise_error(Puppet::Error,/ssh::sshd_config_port must be a valid number and is set to <22invalid>\./)
+    context 'when set to a string' do
+      let (:params) {{'sshd_config_port' => '22222' }}
+
+      it { should contain_file('sshd_config').with_content(/^Port 22222$/) }
+    end
+
+    context 'when not set to a valid number' do
```
Phil-Friderici: a check for valid integers is missing here

andrei693:
Good points, used sshd_config_port_array as variable in template and validate_numeric on that beforehand. Added a test for a valid integer, didn't do a test for an invalid value or string inside an array.

Phil-Friderici:
@andrei693 great work !
@ghoneycutt /me happy

ghoneycutt:
Thanks everyone!!

Released in v3.40.0

PR #163, opened 2016-05-11
add optional selinux support

jlambert121:


jlambert121:
The build failures look like they exist on master as well.

manifests/init.pp
```diff
@@ -737,6 +745,14 @@
     }
   }
 
+  if $::selinux and $manage_selinux_real and $sshd_config_port != '22' {
```
ghoneycutt: Why would we not do this if the port is the default 22/tcp ?

jlambert121: SELinux comes with that label by default so just avoiding the exec on the machine if it's still on port 22.  I could see an argument to just always include it too though - either way.

ghoneycutt:
PR #162 is adding the ability to have multiple ports, so this would need to work with that approach.

Could we use the selmodule type to handle this instead of an exec? I don't think that code around handling selinux should be in the ssh module. I would rather it be elsewhere and the ssh module be able to use it, similar to how the firewall rules are handled.

https://docs.puppet.com/puppet/3.8/reference/types/selmodule.html

jlambert121:
Multiple ports - I'll have to look into that some more.  You want to maintain puppet 3.x compatibility, correct?

I don't think the selmodule type will really work here, but maybe someone can give me more info on how it would.  I considered adding [a selinux](https://forge.puppetlabs.com/jfryman/selinux) module, but was trying to avoid that dependency.  SELinux support in puppet isn't awesome - I've thought adding some native types/providers but just haven't gotten around to it.

manifests/init.pp
ghoneycutt: We don't want to make assumptions on the initial state of the system. It is better to be declarative and say that we want it and the idempotent nature of Puppet will ensure it is there.

ghoneycutt:
Yes, puppet v3 compatibility is required. Perhaps there is another selinux module with types/providers that would work? I'm ok with having a dependency on that as it would be the same as the firewall dependency.

ghoneycutt:

PR #165, opened 2016-05-25
Jthiesfeld maxauthtries

jthiesfeld:
Adding sshd_config_maxauthtries integer variable for sshd_config file.


ghoneycutt:
Are you combining this with the README? Would be glad to merge this functionality. Any new parameters must be documented, validated and tested with the spec tests.

PR #170, opened 2016-06-06
Add param to manage MaxAuthTries in sshd_config

ghoneycutt:


ghoneycutt:
Will be released in v3.38.0 which is being prepared in PR #171

PR #173, opened 2016-06-07
allow ecdsa-sha2-nistp256 hostkeys,  add host_aliases attribute to sshkey resource, add support for PubkeyAuthentication

florianfa:


spec/classes/init_spec.rb
```diff
@@ -472,6 +478,7 @@
     it { should contain_file('sshd_config').with_content(/^AuthorizedKeysCommand \/path\/to\/command$/) }
     it { should contain_file('sshd_config').with_content(/^AuthorizedKeysCommandUser asdf$/) }
     it { should contain_file('sshd_config').with_content(/^HostbasedAuthentication no$/) }
+    it { should contain_file('sshd_config').with_content(/^PubkeyAuthentication yes$/) }
```
ghoneycutt: This should be set to a non-default value; in our case, 'no'.

ghoneycutt:
Thanks for the work!

This should have been three separate PR's because three distinct things are being added, though it is simple enough that I can merge this after the spec test is updated.

florianfa:
I've changed PubkeyAuthentication to no. Next time I try to make different PR's.
Regards Florian

spec/classes/init_spec.rb
florianfa: If I change the PubkeyAuthentication to no the test would break.

florianfa: additionally added param :sshd_pubkeyauthentication  => 'no'

ghoneycutt:
Thank you!

Released in v3.39.0

PR #174, opened 2016-06-08
Add sshd_config options PermitEmptyPasswords and PermitUserEnvironment

jthiesfeld:
Our policy dictates that we explicitly deny the two options even though they default to no.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (C) 2010-2013 Garrett Honeycutt <code@garretthoneycutt.com>
+Copyright (C) 2010-2016 Garrett Honeycutt <code@garretthoneycutt.com>
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.


README.md
```diff
@@ -222,6 +222,18 @@ String or Array to specify address(es) for which sshd will bind. Corresponds to
 
 - *Default*: undef
 
+sshd_config_permitemptypasswords
+--------------------------------
+PermitEmptyPasswords option in sshd_config.  When password authentication is allowed, it specifies whether the server allows login to accounts with empty password strings.
```
ghoneycutt: After the description add this to the line.

`Valid values are 'yes' and 'no'.`

ghoneycutt:
Hi @jthiesfeld 

Glad you are getting involved in sending me PR's :)

Highly suggest you do not work off of your master branch and instead create a branch and then push that branch up and send a PR from it. That way you can work on multiple things without polluting your master branch. 

It also makes rebasing your work off of my master branch easier, which I'll need to do for this before it can be merged.

spec/classes/init_spec.rb
```diff
@@ -473,6 +473,8 @@
     it { should contain_file('sshd_config').with_content(/^HostKey \/etc\/ssh\/ssh_host_rsa_key/) }
     it { should contain_file('sshd_config').with_content(/^HostKey \/etc\/ssh\/ssh_host_dsa_key/) }
     it { should contain_file('sshd_config').with_content(/^StrictModes yes$/) }
+    it { should_not contain_file('sshd_config').with_content(/^PermitUserEnvironment/) }
```
ghoneycutt: These are in the section 'with params used in sshd_config set on valid osfamily' so that means they should be set to a valid, non-default value.

jthiesfeld: Are you saying I should have an entry with the default value in each OS block from lines 25-185?  The current implementation leaves these two options commented out, so I didn't want to add something to the configuration unless it was explicitly stated.

Would adding default values also mean updating the manifest to set the default values?

Do you have a good resource for finding out what the default values are?

ghoneycutt:
This looks good! Please rebase this into a single commit and I will be happy to merge.

PR #179, opened 2016-06-22
Split per-system variables into params.pp as it best practices for mu…

olevole:
…lti-OS modules (Cleanup for root group: some places was hardcoded on 'root' value instead of ssh_config_group variables. Add FreeBSD platform support: tested on FreeBSD 10.x and FreeBSD 11

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -13,7 +13,7 @@ class ssh (
   $ssh_config_hash_known_hosts         = 'USE_DEFAULTS',
   $ssh_config_path                     = '/etc/ssh/ssh_config',
   $ssh_config_owner                    = 'root',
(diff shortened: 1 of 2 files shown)


ghoneycutt:
@olevole 

Thank you very much for sending this! I do not want to adopt the params pattern because it only makes sense when dealing with multiple classes that are not related and need to access the same data.

I would very much like to add the functionality of supporting FreeBSD. Could you please update this PR to add FreeBSD support without adding the params.pp bit? This should also update the README and the metadata.json to specify which versions are supported.

olevole:
Okay, wip

olevole:
done: https://github.com/ghoneycutt/puppet-module-ssh/pull/180

PR #181, opened 2016-06-29
Expose TCPKeepAlive from sshd_config

andrei693:
This will expose TCPKeepAlive parameter from sshd_config via ssh::sshd_config_tcp_keepalive. Defaults to 'yes'.


ghoneycutt:
Thank you!

Released in v3.44.0

PR #185, opened 2016-08-15
WIP - Add support for FreeBSD 10, 11 and 12

ghoneycutt:


manifests/init.pp
```diff
@@ -750,41 +802,57 @@
...
+        ensure  => file,
+        path    => $sshd_config_banner,
+        owner   => $sshd_banner_owner,
+        group   => $sshd_banner_group,
+        mode    => $sshd_banner_mode,
+        content => $sshd_banner_content,
+        require => undef,
+      }
+    } else {
+      file { 'sshd_banner' :
```
ghoneycutt: This I dont like because it violates DRY by having me repeat the resource. The other approach would be to use the inline `?` operator, though that is generally frowned up, though given this case, it may be the correct way to go.

The other approach would be to generate a string for the require, though using a template for some string manipulation in ruby sounds even uglier.

ghoneycutt:
Instead of using 'root' and 'wheel', perhaps we should just use '0' as that will work, since wheel is gid 0 on FreeBSD. This would also remove a bunch of conditional logic.

strangelittlemonkey:
This isn't entirely accurate, though it'd be better than the existing state. There are actually openssh-portable packages for FreeBSD and DragonFly BSD, but they're a separate thing. It would complicate your module somewhat, but the thing is, on those two at least, you should be able to specify on override of wanting it as a package or not. If not, openssh-portable will be in base anyway and build as part of world, the config would be in /etc/, but if it were installed by package, the binaries would be in /usr/local/bin and the configs would be in /usr/local/etc/. With regard to GID/UID by 0, it's generally easier, especially since RHEL based systems are moving to wheel as well while Debian based are sticking to root for some time.

ghoneycutt:

PR #199, opened 2016-10-24
Add option and parameter for TrustedUserCAKeys.

jschrama:
This commit adds the `TrustedUserCAKeys` option to the sshd_config
file as well as a corresponding parameter to the public class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,6 +376,14 @@ String of user for AuthorizedKeysCommandUser in sshd_config.
 
 - *Default*: undef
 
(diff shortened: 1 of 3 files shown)


ghoneycutt:
This looks good, just needs spec tests which I can help with.

jschrama:
Sorry, I did initially forget about that, then started making the spec tests, and then abandoned the whole idea when my colleague made his own openssh module which we are now using. That is why I ended up closing the pull request.

PR #201, opened 2016-11-10
Remove join from validate_absolute_path for sshd_config_hostkey

bc-bjoern:
The validate_absolute_path function is able to parse arrays already. The join results into a invalid absolute path.


ghoneycutt:
Thank you!

PR #203, opened 2016-11-28
define permittunnel to prevent getting locked out

ymajik:
If ssh::sshd_config_permittunnel is not defined, no value gets printed after PermitTunnel, locking you of of an SSH server (just experienced it on a couple of machines at home... :s ). Fixed it with an if loop and default value if not defined.

### DIFF
--- a/templates/sshd_config.erb
+++ b/templates/sshd_config.erb
@@ -194,8 +194,11 @@ MaxSessions <%= @sshd_config_maxsessions %>
 #MaxSessions 10
 <% end -%>
 
-#PermitTunnel no
+<% if @sshd_config_permittunnel -%>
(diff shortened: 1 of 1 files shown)


ghoneycutt:
The default value is already set to 'no'. I don't understand what this is trying to accomplish.

https://github.com/ghoneycutt/puppet-module-ssh#sshd_config_permittunnel

ymajik:
I have to check the condition again, but on several machines I manage with puppet, it did not apply the default value of `PermitTunnel no`, locking me out. I'll see if I can reproduce it.

ymajik:
Apologies for the late answer, I've reverted to your module again instead of my branch, and it does not happen anymore, probably some race condition.

Issue #208, opened 2017-01-20
Solaris 10 config options issue


redno2:
Failed on Solaris 10, the options TCPKeepAlive and PermitTunnel doesn't exist..

-bash-3.2# cat /etc/release     
                    Oracle Solaris 10 1/13 s10x_u11wos_24a X86
  Copyright (c) 1983, 2013, Oracle and/or its affiliates. All rights reserved.
                            Assembled 17 January 2013

/opt/puppetlabs/bin/puppet --version    
4.3.2

[ Jan 20 10:47:40 Executing start method ("/lib/svc/method/sshd start") ]
/etc/ssh/sshd_config: line 105: Bad configuration option: TCPKeepAlive
/etc/ssh/sshd_config: line 119: Bad configuration option: PermitTunnel
/etc/ssh/sshd_config: terminating, 2 bad configuration options

Phil-Friderici:
+1 having seen the same issue here. Hope I can find time soon to create & propose a fix.

Phil-Friderici:
v3.43.0 is the last version that does not break on Solaris.

Phil-Friderici:
Thanks to @nalyanyam I got a list of Solaris incompatible parameters:

KerberosOrLocalPasswd - not valid
KerberosTicketCleanup - not valid
KerberosGetAFSToken - not valid
TCPKeepAlive -  not valid
ShowPatchLevel - not valid
MaxSessions - not valid
PermitTunnel - not valid

We should find a way to exclude them from Solaris SSHd for the future.
Two way come into my mind here:
- Solaris specific template without the incompatible parameters
- enable the template to print out lines only if the value is not undef and set OS default accordingly

I would prefer last one as it could help in other situattions too.
@ghoneycutt What do you think ?

ghoneycutt:
I'd like to keep one template and configure it so that if a setting is undef it will not show up.

ghoneycutt:
Perhaps we could tap into E's Jenkins service and get a few Solaris VM's. We could build a job that runs the code from a PR and tests it on a Solaris VM with serverspec and then reverts the VM. This would give us an automated way to ensure Solaris is good.

Phil-Friderici:
@ghoneycutt Having some Jenkins magic to verify compatibility would be awesome. Getting an environment for full automated testing is on my wish list. Just I believe this will take some time before I can start working into that direction.

Phil-Friderici:
But I will start working on an suggestion for the template.

[...]

PR #213, opened 2017-02-23
Fix parameters not compatible with Solaris

Phil-Friderici:
Suggestion for a fix of issue https://github.com/ghoneycutt/puppet-module-ssh/issues/208

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -806,13 +806,27 @@ describe 'ssh' do
     end
   end
 
-  context 'with ssh_config_hash_known_hosts set to invalid value on valid osfamily' do
-    let(:params) { { :ssh_config_hash_known_hosts => 'invalid' } }
+  describe 'with ssh_config_hash_known_hosts param' do
+    ['yes','no','unset'].each do |value|
+      context "set to #{value}" do
+        let (:params) { { :ssh_config_hash_known_hosts => value } }
 
-    it 'should fail' do
-      expect {
-        should contain_class('ssh')
-      }.to raise_error(Puppet::Error,/ssh::ssh_config_hash_known_hosts may be either \'yes\' or \'no\' and is set to <invalid>\./)
+        if value == 'unset'
+          it { should contain_file('ssh_config').without_content(/^\s*HashKnownHosts/) }
+        else
+          it { should contain_file('ssh_config').with_content(/^\s*HashKnownHosts #{value}$/) }
+        end
+      end
+    end
+
+    context 'when set to an invalid value' do
+      let (:params) { { :ssh_config_hash_known_hosts => 'invalid' } }
+
+      it 'should fail' do
+        expect {
+          should contain_class('ssh')
+        }.to raise_error(Puppet::Error,/ssh::ssh_config_hash_known_hosts may be either \'yes\', \'no\' or \'unset\' and is set to <invalid>\./)
+      end
     end
   end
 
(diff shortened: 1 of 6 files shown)


spec/classes/init_spec.rb
```diff
@@ -971,11 +971,15 @@
...
 
   describe 'sshd_config_permittunnel param' do
-    ['yes','point-to-point','ethernet','no'].each do |value|
+    ['yes','point-to-point','ethernet','no','unset'].each do |value|
       context "set to #{value}" do
         let (:params) { { :sshd_config_permittunnel => value } }
 
-        it { should contain_file('sshd_config').with_content(/^PermitTunnel #{value}$/) }
+        if value == 'unset'
+          it { should contain_file('sshd_config').without_content(/^PermitTunnel/) }
```
ghoneycutt: negative matches should have `/^\s*` as the start of the regex to protect us from whitespace shifting in the template.

templates/sshd_config.erb
```diff
@@ -166,7 +166,9 @@ X11UseLocalhost <%= @sshd_x11_use_localhost %>
 PrintMotd <%= @sshd_config_print_motd %>
 #PrintLastLog yes
 #TCPKeepAlive yes
-TCPKeepAlive <%= @sshd_config_tcp_keepalive %>
+<% if @sshd_config_tcp_keepalive_real =~ /(yes|no)/ -%>
```
ghoneycutt: Would `!= 'unset'` make this easier to change in the future as valid options to the parameter change?

templates/sshd_config.erb
```diff
@@ -195,7 +197,9 @@ MaxSessions <%= @sshd_config_maxsessions %>
 <% end -%>
 
 #PermitTunnel no
-PermitTunnel <%= @sshd_config_permittunnel %>
+<% if @sshd_config_permittunnel_real =~ /(yes|no|point-to-point|ethernet)/ -%>
```
ghoneycutt: Would `!= 'unset'` make this easier to change in the future as valid options to the parameter change?

ghoneycutt:
I'm taking it that the other parameters in #208 are not actual issues.

templates/sshd_config.erb
Phil-Friderici: good idea

Phil-Friderici:
After your review, I'll add the same logic for the other parameters mentioned in https://github.com/ghoneycutt/puppet-module-ssh/issues/208

Phil-Friderici:
This change does introduce the possibility to use 'unset' as new value for all incompatible Solaris parameters mentioned in https://github.com/ghoneycutt/puppet-module-ssh/issues/208 that are changeable.

@ghoneycutt It does not take care for unparameterized parameters yet. Should I add a note with the other incompatible parameters somewhere to avoid the problems in the future ?

Let me know when I should squash the commit into one.

ghoneycutt:
Documenting the items sure can't hurt and would be appreciated. Could you please add that and squash your commits and I'll get this merged.

Phil-Friderici:
This change does introduce the possibility to use 'unset' as new value for all incompatible Solaris parameters mentioned in https://github.com/ghoneycutt/puppet-module-ssh/issues/208 that are changeable.

@ghoneycutt It does not take care for unparameterized parameters yet. Should I add a note with the other incompatible parameters somewhere to avoid the problems in the future ?

Let me know when I should squash the commit into one.

Phil-Friderici:
Squashed and documented.

Asked @nalyanyam to test the code.
Let's wait for his feedback.

Phil-Friderici:
@nalyanyam did reply positivly:
> Tested the updated ssh module on Solaris and all ok, Works fine by default or with ‘unset’ values.

Please feel free to merge ;)

ghoneycutt:
Thanks @Phil-Friderici @nalyanyam and @redno2

Released in v3.49.1

Phil-Friderici:
Big THANKS to all of you gentlemen :)

PR #215, opened 2017-02-28
Make sure the ssh config dir is present before trying to define the known host file.

Arakmar:
An error occurs with a fresh install (CentOS 7 in this case) when the SSH package isn't installed by default.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -927,11 +927,12 @@ class ssh (
   }
 
   file { 'ssh_known_hosts':
-    ensure => file,
-    path   => $ssh_config_global_known_hosts_file,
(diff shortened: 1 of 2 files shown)


ghoneycutt:
Instead of requiring the ssh_config, shouldn't it require the ssh package?

`require => Package[$packages_real],`

Arakmar:
Yes you're right, I thought it was the /etc/ssh dir...
It's updated.

Arakmar:
Sorry for the very late reply...
I added the require in the spec tests.

ghoneycutt:
Thank you!! Released in v3.51.1

PR #216, opened 2017-03-03
Option to not manage global_known_hosts

kTitan:
I have here currently a special setup, in which I need to manage the global_known_hosts_file (/etc/ssh/ssh_known_hosts) with an other puppet module.
But I still want use this great module, but I do not see an option available here to ignore that file. Because of that I got a conflict.

Did I have overlooked an option for that?
If there is no option available for that I see currently to possible ways to achive that:
1. If ssh_key_import and purge_keys are the to false, simply ignore the file (I found no other code part that puts content into that file)
2. Create a new option manage_global_known_hosts_file

What do you think about that? If you tell me your prefered way, I can create a PR for that if you want.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -111,6 +111,7 @@ class ssh (
   $root_ssh_config_content             = "# This file is being maintained by Puppet.\n# DO NOT EDIT\n",
   $sshd_config_tcp_keepalive           = undef,
   $sshd_config_permittunnel            = undef,
+  $manage_global_known_hosts           = true,
 ) {
 
   case $::osfamily {
@@ -763,6 +764,13 @@ class ssh (
   #ssh_config template
   validate_string($ssh_config_template)
 
+  if type3x($manage_global_known_hosts) == 'string' {
+    $manage_global_known_hosts_real = str2bool($manage_global_known_hosts)
+  } else {
+    $manage_global_known_hosts_real = $manage_global_known_hosts
+  }
+  validate_bool($manage_global_known_hosts_real)
+
   #sshd_config template
   validate_string($sshd_config_template)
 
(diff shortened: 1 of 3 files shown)


kTitan:
Here is my try for the PR.
But I am not sure if I got the spec tests correctly. Feel free to comment, what I should change

spec/classes/init_spec.rb
```diff
@@ -1962,6 +1973,45 @@
...
+            should contain_class('ssh')
+          }.to raise_error(Puppet::Error,/(is not a boolean|Unknown type of boolean)/)
+        end
+      end
+    end
+
+    ['true', true].each do |value|
+      context "specified as valid true value #{value} (as #{value.class})" do
+        let(:params) { { :manage_global_known_hosts => value } }
+        let(:params) { { :ssh_config_global_known_hosts_file => '/valid/path' } }
```
ghoneycutt: since we are testing the values of manage_global_known_hosts and not ssh_config_global_known_hosts_file, ssh_config_global_known_hosts_file should not be specified. Please remove that line and use the default path instead of `/valid/path`

spec/classes/init_spec.rb
```diff
@@ -295,6 +295,7 @@
   context 'with default params on invalid osfamily' do
     let(:facts) { default_facts.merge({ :osfamily => 'C64' }) }
     let(:params) { { :manage_root_ssh_config => 'invalid' } }
+    let(:params) { { :manage_global_known_hosts => 'invalid' } }
```
ghoneycutt: Please remove this line.

manifests/init.pp
```diff
@@ -898,24 +906,28 @@
...
+    # remove ssh key's not managed by puppet
+    resources  { 'sshkey':
+      purge => $purge_keys_real,
     }
-  }
 
-  # remove ssh key's not managed by puppet
-  resources  { 'sshkey':
-    purge => $purge_keys_real,
   }
```
ghoneycutt: This is trivial, but since you are already rebasing changes for the tweaks to the spec tests, could you please remove the empty line between closing brackets.

So instead of

```puppet
  # remove ssh key's not managed by puppet
  resources  { 'sshkey':
    purge => $purge_keys_real,
  }

}
```

put

```puppet
  # remove ssh key's not managed by puppet
  resources  { 'sshkey':
    purge => $purge_keys_real,
  }
}
```

kTitan:
I have updated the PR based on your requested changes. 
Thanks for the comments :)

kTitan:
Would be nice to get a feedback, if the PR is now ok, and will be pushed into one of the next releases :)

ghoneycutt:
@kTitan Looks good, though I'd like to get another set of eyes on this. @Phil-Friderici could you please take a look.

Phil-Friderici:
When ```$manage_global_known_hosts``` is set to false it also disables ```resources{sshkey}``` and ```Sshkey <<||>>```.
Would be good to see this change of functionality in the specs tests too. But I have to confess that I don't know how to test ```Sshkey``` here myself.

PR #217, opened 2017-03-06
Add support for 'AuthenticationMethods'

Saford91:
I use this option to require both password and public key authentication on CentOS 7. This option does not work for the version of OpenSSH packaged for CentOS 6.

### DIFF
--- a/templates/sshd_config.erb
+++ b/templates/sshd_config.erb
@@ -196,6 +196,10 @@ MaxSessions <%= @sshd_config_maxsessions_integer %>
 #MaxSessions 10
 <% end -%>
 
+<% if @sshd_config_authenticationmethods != [] -%>
+AuthenticationMethods  <%= @sshd_config_authenticationmethods.join(' ') -%>
+<% end -%>
+
 #PermitTunnel no
 <% if @sshd_config_permittunnel_real != nil -%>
 PermitTunnel <%= @sshd_config_permittunnel_real %>
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -52,6 +52,7 @@ class ssh (
   $sshd_config_ciphers                 = undef,
   $sshd_config_kexalgorithms           = undef,
   $sshd_config_macs                    = undef,
+  $sshd_config_authenticationmethods   = [],
   $ssh_enable_ssh_keysign              = undef,
   $sshd_config_allowgroups             = [],
   $sshd_config_allowusers              = [],


ghoneycutt:
Hi @Saford91 

Thank you for the contribution. Additional parameters need to have accompanying spec tests and be documented in the README.

templates/sshd_config.erb
```diff
@@ -196,6 +196,10 @@ MaxSessions <%= @sshd_config_maxsessions_integer %>
 #MaxSessions 10
 <% end -%>
 
+<% if @sshd_config_authenticationmethods != [] -%>
+AuthenticationMethods  <%= @sshd_config_authenticationmethods.join(' ') -%>
```
doomnuggets: The join character should be `,` instead of a space.
Quoting the man page of sshd(5)

>      AuthenticationMethods
> 	     Specifies the authentication methods that must be successfully
> 	     completed for a user to be	granted	access.	 This option must be
> 	     followed by one or	more comma-separated lists of authentication
> 	     method names.  Successful authentication requires completion of
> 	     every method in at	least one of these lists.
> 
> 	     For example, an argument of ``publickey,password
> 	     publickey,keyboard-interactive'' would require the	user to	com-
> 	     plete public key authentication, followed by either password or
> 	     keyboard interactive authentication.  Only	methods	that are next
> 	     in	one or more lists are offered at each stage, so	for this exam-
> 	     ple, it would not be possible to attempt password or keyboard-
> 	     interactive authentication	before public key.
> 
> 	     For keyboard interactive authentication it	is also	possible to
> 	     restrict authentication to	a specific device by appending a colon
> 	     followed by the device identifier ``bsdauth'', ``pam'', or
> 	     ``skey'', depending on the	server configuration.  For example,
> 	     ``keyboard-interactive:bsdauth'' would restrict keyboard interac-
> 	     tive authentication to the	``bsdauth'' device.
> 
> 	     If	the ``publickey'' method is listed more	than once, sshd(8)
> 	     verifies that keys	that have been used successfully are not
> 	     reused for	subsequent authentications.  For example, an
> 	     AuthenticationMethods of ``publickey,publickey'' will require
> 	     successful	authentication using two different public keys.
> 
> 	     This option will yield a fatal error if enabled if	protocol 1 is
> 	     also enabled.  Note that each authentication method listed	should
> 	     also be explicitly	enabled	in the configuration.  The default is
> 	     not to require multiple authentication; successful	completion of
> 	     a single authentication method is sufficient.

Saford91: Yes and no. It accepts more than one comma separated list.
```
AuthenticationMethods  publickey,password publickey,gssapi-with-mic
```
An array of arrays could work here, but I'm not sure how to make that print in an ERB.

baurmatt:
Also interested in this to get https://wiki.mozilla.org/Security/Guidelines/OpenSSH#Modern_.28OpenSSH_6.7.2B.29 working. Are you going to fix the PR? If not, i'm happy to finish it.

Saford91:
I'm not too familiar with writing unit tests, so if you're OK with
finishing it, go ahead.

ghoneycutt:
Support released in v3.57.0

Thanks for your help!!

PR #223, opened 2017-05-16
Add sshd_config_hostcertificate and sshd_config_trustedusercakeys parameters

### LINKED ISSUE #221: Allow management of the HostCertificate and TrustedUserCAKeys
We need the ability to manage the `HostCertificate` option for the SSH Daemon.

Suggestion:

`sshd_host_certificate` parameter to set the `HostCertificate` option. The value would be a `String`, or in Puppet 4 Data Types, an `Stdlib::Absolutepath` type.

`sshd_user_trusted_ca` parameter to set the `TrustedUserCAKeys` option. The value would be a `String` or an `Array` of `String`s. In Puppet 4 Data Types, that would be a `Stdlib::Absolutepath` or an `Array` of `Stdlib::Absolutepath` data types.

dhollinger:
closes #221

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -112,6 +112,8 @@ class ssh (
   $root_ssh_config_content             = "# This file is being maintained by Puppet.\n# DO NOT EDIT\n",
   $sshd_config_tcp_keepalive           = undef,
   $sshd_config_permittunnel            = undef,
+  $sshd_config_hostcertificate         = undef,
+  $sshd_config_trustedusercakeys       = undef,
 ) {
 
   case $::osfamily {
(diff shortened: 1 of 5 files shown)


dhollinger:
Added Work in Process [WIP] to title while I get the test issues figured out.

dhollinger:
@ghoneycutt Is there something I'm missing in how your tests are setup?

Any test I add to the init_spec file causes the same 15 tests to fail:

13 tests checking the sshd_content against the files in fixtures and 2 regex content checks later in the file.

dhollinger:
@alexjfisher Nice catch. Didn't expect that to be the issue.

ghoneycutt:
This PR looks great!

templates/sshd_config.erb
```diff
@@ -254,3 +254,13 @@ Match <%= key %>
 <%     end -%>
 <%   end -%>
 <% end -%>
+
```
ghoneycutt: These spaces are causing the issue since it doesn't match the fixture file. Put your opening `if` just after the last `end` and then put an empty line. Same with the if below it.

dhollinger: Thanks! I have updated the template and tests are passing.

manifests/init.pp
```diff
@@ -813,6 +825,14 @@
     validate_re($sshd_config_permittunnel_real, '^(yes|no|point-to-point|ethernet|unset)$', "ssh::sshd_config_permittunnel may be either 'yes', 'point-to-point', 'ethernet', 'no' or 'unset' and is set to <${sshd_config_permittunnel_real}>.")
   }
 
+  if $sshd_config_hostcertificate_real != undef {
+    validate_absolute_path($sshd_config_hostcertificate_real)
+  }
+
+  if $sshd_config_trustedusercakeys_real != undef {
+    validate_absolute_path($sshd_config_trustedusercakeys_real)
```
ghoneycutt: https://www.freebsd.org/cgi/man.cgi?sshd_config(5)

This can also be `'none'`.

ghoneycutt:
Thanks everyone! I've squashed these commits in PR #224 along with some minor changes and will merge and cut a new release once TravisCI is satisfied.

PR #224, opened 2017-05-17
223 new params

ghoneycutt:


ghoneycutt:
Thanks @dhollinger 

Released in v3.51.0

PR #230, opened 2017-06-23
Added possibility to add AcceptEnv/SendEnv custom strings

apazga:
Added `sshd_acceptenv_string` and `ssh_sendenv_string` params, to add custom AcceptEnv/SendEnv strings.
Some software need another parameters, like `LANG LC_*`, (like Proxmox VE: https://forum.proxmox.com/threads/pve-4-3-console-access-on-different-cluster-member.29733/page-2 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,12 @@ Boolean to enable SendEnv options for specifying environment variables. Default 
 
 - *Default*: 'USE_DEFAULTS'
 
+ssh_sendenv_string
+-------------
+String for SendEnv option for specifying custom environment variables.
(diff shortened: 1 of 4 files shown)


flhe:
We need a dynamic AcceptEnv too. Is it possible to integrate it ?

Phil-Friderici:
In v4 you can specify what ever is needed for `AcceptEnv` and `SendEnv` via `$ssh::server::accept_env` and `$ssh::send_env`.
Please re-open if this doesn't fit your use case.

PR #240, opened 2017-07-21
Add more SSH CA related fixes and paramaters

dhollinger:
#### Enhancements
* Add `$sshd_config_authorized_principals_file`. This sets the value of the SSH `AuthorizedPrincipalsFile` setting.
* Update/Add tests for all changes
* Update README

#### Bug Fixes
* Update `$sshd_config_hostcertificate` to take an array of values. The original implementation was bugged. `HostCertificate` must be tied to a `HostKey` 1:1.


dhollinger:
There were so many jobs in the Travis matrix that the tests were failing to even complete, let alone pass in some cases on older jobs. To simplify tests, without removing "support", I have removed all unsupported versions of Puppet and their associated Ruby versions from the matrix.

We can begin adding tests for the EOL versions of Puppet back in, one at a time, to hopefully prevent further issues with travis, but I was just trying to get the tests to at least run again.

.travis.yml
```diff
@@ -2,34 +2,17 @@
 language: ruby
 
 rvm:
```
ghoneycutt: please revert changes to Travis. The testing has now been fixed from PR #241

README.md
```diff
@@ -597,7 +597,7 @@ ssh::sshd_config_match:
 
 sshd_config_hostcertificate
 ---------------------------
-Absolute path to the OpenSSH Host CA Certificate (HostCertificate) for use with SSH CA validation for Host Certificates.
+An Array of Absolute Paths to the Host CA Public Key. Each entry *MUST* be tied 1:1 to a Host CA Private Key (see [sshd_config_hostkey](#sshd_config_hostkey))
```
ghoneycutt: This should support a single value given as a string for backward compatibility.

dhollinger: Does that matter if the parameter was incorrect in the first place?

I'm fixing it as the original parameter was "broken" already from the SSH standpoint. Meaning the value could only support one HostKey + HostCertificate combination, when in reality each HostCertificate should be directly tied to a HostKey?

So from a Puppet standpoint, this is a breaking change, but from an SSH standpoint this is a bugfix

README.md
```diff
@@ -598,7 +598,7 @@ ssh::sshd_config_match:
 
 sshd_config_hostcertificate
 ---------------------------
-Absolute path to the OpenSSH Host CA Certificate (HostCertificate) for use with SSH CA validation for Host Certificates.
+An Absolute Path or Array of Absolute Paths orto the Host CA Public Key. Each entry *MUST* be tied 1:1 to a Host CA Private Key (see [sshd_config_hostkey](#sshd_config_hostkey))
```
ghoneycutt: typo 'orto'

README.md
```diff
@@ -608,6 +608,14 @@ Absolute path to the OpenSSH User CA Certificate (TrustedUserCAKeys) for use wit
 
 - *Default*: undefined
 
+sshd_config_authorized_principals_file
+--------------------------------------
+String path (relative or absolute) to the `authorized_principals` file. Sets the `AuthorizedPrincipalsFile` setting in `sshd_config`
```
ghoneycutt: Should we just impose absolute paths here? It will be easier to validate and is probably the right thing to do if there is an option.

dhollinger: No, because the default setting is not an absolute path and users may choose to not use an absolute path. Not to mention that value can also take tokens (which for our purposes can just be part of the string)

ghoneycutt:
Big thanks for this!

Released in v3.54.0

PR #245, opened 2017-09-15
Add ssh::config_entry defined type

jeffmccune:
This patch adds a mechanism to manage ~/.ssh/config entries using the
concat module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@ ssh_key_ensure and purge_keys.
 
 This module may be used with a simple `include ::ssh`
 
+The `ssh::config_entry` defined type may be used directly and is used to manage
+Host entries in a personal `~/.ssh/config` file.
+
 ===
 
 ### Table of Contents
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -8,6 +8,9 @@ ssh_key_ensure and purge_keys.
 
 This module may be used with a simple `include ::ssh`
 
+The `ssh::config_entry` defined type may be used directly and is used to manage
+Host entries in a personal `~/.ssh/config` file.
+
```
ghoneycutt: @Phil-Friderici We could use a param `ssh::config_entries` that is a hash that is used with create_resources for ssh::config_entry resources.

manifests/config_entry.pp
```diff
@@ -0,0 +1,36 @@
+# == Define: ssh::config_entry
+#
+# Manage an entry in ~/.ssh/config for a particular user.  Lines model the lines
+# in each Host block.
+define ssh::config_entry (
```
ghoneycutt: @Phil-Friderici we could also use spec tests. This is for our friends in NY.

README.md
Phil-Friderici: sure, like in the types and other modules

manifests/config_entry.pp
Phil-Friderici: will be my pleasure

PR #246, opened 2017-09-26
Add ssh::config_entry defined type (with ssh::config_entries and tests)

Phil-Friderici:


Phil-Friderici:
Includes/Fixes #245 

Adds param ssh::config_entries, a hash that is used with create_resources for ssh::config_entry resources.
Adds spec tests.

ghoneycutt:
Thanks!

Released in v3.55.0

Issue #244, opened 2017-09-15
ServerKeyBits Deprecated in RHEL 7.4


spacezer0:
As of RHEL 7.4 ServerKeyBits is now a deprecated option.

Warning Message in /var/log/messages:
Deprecated option ServerKeyBits

smokris:
The `ServerKeyBits` option [only applies to SSH protocol version 1](https://man.openbsd.org/OpenBSD-6.0/sshd_config).  Support for that was [removed in OpenSSH version 7.6](https://www.openssh.com/txt/release-7.6), so this issue occurs not just on RHEL but on any system running OpenSSH 7.6 or later.  PR #259 removes that option from the Puppet-generated `sshd_config`.

westbywest:
In addition, having this parameter take the string 'USE_DEFAULTS' as its default value prevents one from setting the param to undef to deactivate the line in sshd_config.  I.e. the module can't be made to not insert a ServerKeyBits value into sshd_config.

irekpo:
Hi Guys,
I don't think this problem is fixed. I am using the latest Centos (7.5.1804). I was using previous version of you module when I have noticed the problem. However after the upgrade ghoneycutt-ssh (v3.54.0 -> v3.57.0) the problem stays. I need to modify your code to fix that. Did I missed some steps or the code needs some correction?

mears148:
This is the same for RhostsRSAAuthentication...also a deprecated option as of 7.4.

mears148:
I'd make the change but I have little free time at the moment.

PR #259, opened 2018-01-22
Remove the obsolete ServerKeyBits sshd_config option

smokris:
#244

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,7 +49,6 @@ class ssh (
   $sshd_config_use_dns                    = 'USE_DEFAULTS',
   $sshd_config_authkey_location           = undef,
   $sshd_config_strictmodes                = undef,
-  $sshd_config_serverkeybits              = 'USE_DEFAULTS',
   $sshd_config_banner                     = 'none',
   $sshd_config_ciphers                    = undef,
   $sshd_config_kexalgorithms              = undef,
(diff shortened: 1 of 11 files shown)


ghoneycutt:
Hi @smokris 

Thank you for your contribution. We can't just remove this option, because it is a valid option and could be in use and is used by default on some of the older platforms. If there are specific platforms in which it should not be used, such as EL7, I would be happy to merge changes to turn it off there.

manifests/init.pp
```diff
@@ -143,7 +142,6 @@
       $default_sshd_gssapicleanupcredentials   = 'yes'
       $default_sshd_acceptenv                  = true
       $default_service_hasstatus               = true
-      $default_sshd_config_serverkeybits       = '1024'
```
ghoneycutt: Needs a conditional here so that if EL7 then set to `undef`, else set to `'1024'`.

tuxmea:
Fix released in https://github.com/ghoneycutt/puppet-module-ssh/pull/278 this one can be closed.

ghoneycutt:
Fix is in v3.57.1

EpiqSty:
Looks like that PR was'nt merged, but we are observing such errors on Debian 9 systems:

` rexec line 34: Deprecated option ServerKeyBits`

So, my question - is there any way to disable ServerKeyBits for given distribution via Hiera, or we have to hardcode every OS using the same way as it was done for RHEL 7.4 (#278)? 

@ghoneycutt , @tuxmea , @smokris

PR #262, opened 2018-02-16
make ssh key export conditional with new boolean ssh_key_export

### LINKED ISSUE #229: Warning of exported resources
If using this module on a system without exported resources (which is reasonably common, and the only option for masterless setups), then you get two warnings:
```
Warning: You cannot collect exported resources without storeconfigs being set; the export is ignored at /srv/maverick/software/maverick/manifests/puppet-modules/ssh/manifests/init.pp:935:3
Warning: You cannot collect exported resources without storeconfigs being set; the collection will be ignored at /srv/maverick/software/maverick/manifests/puppet-modules/ssh/manifests/init.pp:953:5
```
The second can be suppressed by setting:
`ssh_key_import              => false`

The export however is unconditional.  It should be wrapped by an option so it can be turned off, and the warning suppressed.

henry-spanka:
Currently a warning is displayed on systems that do not support exported resources (have no PuppetDB).
This pull request adds an option that allows the user to disable exporting the ssh key so the warning is not displayed.

For backwards compatibility the new parameter `ssh_key_export`is set to `true` by default which enables exporting the ssh key.

Closes #229, #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,7 +423,7 @@ X11Forwarding in sshd_config. Specifies whether X11 forwarding is permitted.
 
 sshd_x11_use_localhost
 ----------------------
-X11UseLocalhost in sshd_config. Specifies if sshd should bind the X11 forwarding server 
+X11UseLocalhost in sshd_config. Specifies if sshd should bind the X11 forwarding server
 to the loopback address or to the wildcard address.
 
 - *Default*: 'yes'
@@ -792,9 +792,15 @@ Boolean to declare whether the service's init script has a functional status com
 
 ssh_key_ensure
 --------------
-Export node SSH key. Valid values are 'present' and 'absent'.
+Export node SSH key. Valid values are 'present' and 'absent'. Requires ssh_key_export set to 'true'.
 
 - *Default*: 'present'
+
+ssh_key_export
+--------------
+Manage node SSH key with exported resources. Valid values are 'true' and 'false'.
+
+- *Default*: 'true'
 
 ssh_key_import
 --------------
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -102,6 +102,7 @@ class ssh (
   $service_hasrestart                     = true,
   $service_hasstatus                      = 'USE_DEFAULTS',
   $ssh_key_ensure                         = 'present',
+  $ssh_key_export                         = true,
   $ssh_key_import                         = true,
   $ssh_key_type                           = 'ssh-rsa',
   $ssh_config_global_known_hosts_file     = '/etc/ssh/ssh_known_hosts',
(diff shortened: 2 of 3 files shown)


spec/classes/init_spec.rb
```diff
@@ -2335,6 +2335,50 @@
...
+      end
+    end
+
+    ['false',false].each do |value|
+      context "as #{value}" do
+        let(:params) { { :ssh_key_export => value } }
+
+        it { should compile.with_all_deps }
+
+        it { should contain_class('ssh') }
```
ghoneycutt: When false, it should check that the sshkey is not present.

```ruby
it { should_not contain_resources('sshkey') }
```

manifests/init.pp
```diff
@@ -102,6 +102,7 @@
   $service_hasrestart                     = true,
   $service_hasstatus                      = 'USE_DEFAULTS',
   $ssh_key_ensure                         = 'present',
+  $ssh_key_export                         = true,
```
ghoneycutt: This needs to be added to the README

ghoneycutt:
Awesome pull request! Thank you very much!! Just needs a few small changes and this can be merged.

spec/classes/init_spec.rb
henry-spanka: I already tried that but the resource is still in the catalog, just not exported.

spec/classes/init_spec.rb
```diff
@@ -2375,6 +2375,12 @@
         it { should compile.with_all_deps }
 
         it { should contain_class('ssh') }
```
henry-spanka: wrapped this in an exported_resource context so the test does not fail as it seems that the sshkey resource is always present just not exported if we set the `ssh_key_export` to false.

README.md
```diff
@@ -792,10 +792,16 @@ Boolean to declare whether the service's init script has a functional status com
...
 ssh_key_ensure
 --------------
-Export node SSH key. Valid values are 'present' and 'absent'.
+Export node SSH key. Valid values are 'present' and 'absent'. Requires ssh_key_export set to 'true'.
 
 - *Default*: 'present'
 
+ssh_key_export
+--------------
+Manage node SSH key with exported resources. Valid values are 'true' and 'false'.
```
henry-spanka: ssh_key_export does nearly the same as ssh_key_ensure. Any suggestion on how to better clarify the difference between them?

spec/classes/init_spec.rb
```diff
@@ -2375,6 +2375,12 @@
...
+
+        context "as an exported resource" do
+            subject { exported_resources }
+
+            it { should_not contain_resources('sshkey') }
```
henry-spanka: wrapped this in an exported_resource context so the test does not fail as it seems that the sshkey resource is always present just not exported if we set the ssh_key_export to false.

spec/classes/init_spec.rb
henry-spanka: Should be done.

henry-spanka:
@ghoneycutt Can we get this merged, thanks?

fnoop:
@henry-spanka many thanks for this.  @ghoneycutt could this please be merged?

ghoneycutt:

PR #265, opened 2018-03-08
Correct obsolete ServerKeyBits sshd_config option on CentOS 7.

jamescaldow:
This commit fixes https://github.com/ghoneycutt/puppet-module-ssh/issues/244.

This is achieved using the creator's advice of using a conditional. No further changes to the code are made.


jamescaldow:
Struggling to correct the spec tests. Code builds and runs fine, but spec tests are failing due to the dependence on the entry being present in all configs. Currently working on trying to resolve the failing tests, but this is my first time submitting a pull request, and my first time working with RSpec.

PR #266, opened 2018-03-28
Fixing an inaccuracy within README.md

zyamada:
README.md incorrectly lists the parameter for configuring the PrintLastLog sshd_config option as "sshd_config_print_lastlog" when in actuality the parameter is, "sshd_config_print_last_log"


ghoneycutt:
Thanks @zyamada

PR #270, opened 2018-04-30
added parameter sshd_config_subsystem_sftp_options for https://access…

depohlm:
added parameter sshd_config_subsystem_sftp_options (https://access.redhat.com/articles/1374633)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -322,6 +322,10 @@ class ssh (
     $sshd_config_subsystem_sftp_real = $sshd_config_subsystem_sftp
   }
 
+  if $sshd_config_subsystem_sftp_options != undef {
+    validate_string($sshd_config_subsystem_sftp_options)
+  }
+
   if $sshd_config_mode    == 'USE_DEFAULTS' {
     $sshd_config_mode_real = $default_sshd_config_mode
   } else {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -73,6 +73,7 @@ class ssh (
   $sshd_banner_mode                       = '0644',
   $sshd_config_xauth_location             = 'USE_DEFAULTS',
   $sshd_config_subsystem_sftp             = 'USE_DEFAULTS',
+  $sshd_config_subsystem_sftp_options     = undef,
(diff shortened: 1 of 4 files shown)


templates/sshd_config.erb
```diff
@@ -238,7 +238,7 @@ XAuthLocation <%= @sshd_config_xauth_location_real %>
 
 <% end -%>
 # override default of no subsystems
-Subsystem sftp <%= @sshd_config_subsystem_sftp_real %>
+Subsystem sftp <%= @sshd_config_subsystem_sftp_real %> <%= @sshd_config_subsystem_sftp_options %>
```
ghoneycutt: the trailing whitespace is causing an issue with the spec tests. You can enclose it and the part you added in an if statement as shown above.

manifests/init.pp
```diff
@@ -73,6 +73,7 @@
   $sshd_banner_mode                       = '0644',
   $sshd_config_xauth_location             = 'USE_DEFAULTS',
   $sshd_config_subsystem_sftp             = 'USE_DEFAULTS',
+  $sshd_config_subsystem_sftp_options     = undef,
```
ghoneycutt: This new parameter needs a spec test associated with it.

ghoneycutt:
Thank you! That should solve the template issue, though we need spec tests added before this can be merged.

manifests/init.pp
```diff
@@ -321,6 +322,10 @@
     $sshd_config_subsystem_sftp_real = $sshd_config_subsystem_sftp
   }
 
+  if $sshd_config_subsystem_sftp_options != undef {
```
depohlm: Spec added here.

ghoneycutt: That's the validation piece. Check out this commit to see how spec tests were added.

https://github.com/ghoneycutt/puppet-module-ssh/commit/bc4c8a82a64c41c2d146b038cbc3e8d6a4fc932b

depohlm:
Spec test added.

Phil-Friderici:
v4 does allow to set `Subsystem` in sshd_config with the `$ssh::server::subsystem` parameter.

PR #280, opened 2018-08-07
Add Identifyfile management (fixes #276)

### LINKED ISSUE #276: add id_ed25519 support
Given: OpenSSH_7.4p1, OpenSSL 1.0.2k-fips on Centos 7 and Puppet 5 and the manifest ```include ssh```
When I: try to use a pre-existing ssh ed25529 key to ssh into another server.
I get: a password prompt.
I expected: to log straight in with the private key.

Investigation shows that this module has wiped out the ssh_config entries as follows:
```
-   IdentityFile ~/.ssh/id_ecdsa
-   IdentityFile ~/.ssh/id_ed25519
```
and the sshd_config entries as follows:
```
-HostKey /etc/ssh/ssh_host_ecdsa_key
-HostKey /etc/ssh/ssh_host_ed25519_key
```
This is a nuisance because we've adopted ed25519 widely already.

Double-checking this morning, SSH with the -v option gave:
```
debug1: permanently_set_uid: 0/0
debug1: key_load_public: No such file or directory
debug1: identity file /root/.ssh/id_rsa type -1
debug1: key_load_public: No such file or directory
debug1: identity file /root/.ssh/id_rsa-cert type -1
debug1: key_load_public: No such file or directory
debug1: identity file /root/.ssh/id_dsa type -1
debug1: key_load_public: No such file or directory
debug1: identity file /root/.ssh/id_dsa-cert type -1
debug1: Enabling compatibility mode for protocol 2.0
```
so ssh is definitely no longer even trying to load the id_ed25519 key.

threepistons:
This defaults to putting in id_rsa and id_dsa default identity files as it did before, so it shouldn't break existing installations of this module.  Suggestion for next release: add sensible defaults for newer OSes.  I added tests for this in the big spec file.

Apply with Hiera
```
ssh::ssh_config_identityfiles:
  - ~/.ssh/id_ed25519
  - ~/.ssh/id_rsa
  - ~/.ssh/id_ecdsa
```
or Puppet
```
class ssh {
  ssh_config_identityfiles => [ '~/.ssh/id_ed25519',
    '~/.ssh/id_rsa',
    '~/.ssh/id_ecdsa'
  ]
}
```
# Documentation
The readme file now has the *ssh_config_identityfiles* parameter documented with an example.


hdep:
seems good I had the issue for #309  @ghoneycutt what do you think ?

clevelas:
I think this needs to be implemented.  RedHat/CentOS 7 use ecdsa and ed25519.  It would be good to support those identity files.

g2payne:
Will this feature or similar ever be implemented? Being able to include ecdsa and ed25519 as IdentityFile options or set to the defaults which includes these really needs to be added.

raoulbhatia:
I agree that the configuration files could be updated to be more modern and/or configuration options could be improved, especially w.r.t. key management.

@g2payne as a work around, you can set `ssh::ssh_config_template` to point to your own template where you could either

* Remove all `IdentityFile` stances (and other configuration options) to revert back to the defaults, or
* Explicitly the configuration you need for your environment.

Check https://github.com/ghoneycutt/puppet-module-ssh/blob/f2b6c2d/templates/ssh_config.erb#L50 ff

PR #281, opened 2018-08-07
Add support for prohibit-password

threepistons:
```PermitRootLogin prohibit-password``` is now permitted by the validator.  No spec test, there wasn't one for the existing validator.  Works on a vm.    Non-breaking change.


bwoods8:
prohibit-password will be needed soon as without-password is deprecated

PR #282, opened 2018-08-07
Update to support ed25519

threepistons:
There's a host key validator that didn't have a case for ed25519.  It now does.  No spec test (there wasn't for the rest of the case block) but it has been tested on a vm and works.  Non-breaking change.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -741,10 +741,13 @@ class ssh (
       $key = $::sshdsakey
     }
     'ecdsa-sha2-nistp256': {
-          $key = $::sshecdsakey
+      $key = $::sshecdsakey
+    }
+    'ssh-ed25519': {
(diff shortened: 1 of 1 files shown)


threepistons:
I don't think that the failing Travis build is caused by my change.
```
Installing rspec-support 3.8.0
RuntimeError: Digest::Base cannot be directly inherited in Ruby
An error occurred while installing rake (12.3.1), and Bundler
cannot continue.
Make sure that `gem install rake -v '12.3.1' --source 'https://rubygems.org/'`
succeeds before bundling.
In Gemfile:
  puppetlabs_spec_helper was resolved to 2.9.1, which depends on
    puppet-syntax was resolved to 2.4.1, which depends on
      rake
```

I didn't change any of the rake stuff.

Issue #288, opened 2018-09-12
Not able to set a custom AcceptEnv to sshd.


renato-farias:

renato-farias:
How can I submit the code for review.

Phil-Friderici:
Closing as `AcceptEnv `can be configured freely with `$ssh::server:accept_env` in v4.
Please re-open if this doesn't fit your use case.

PR #289, opened 2018-09-13
module able to accept custom AcceptEnv to sshd

renato-farias:
Fixing issue #288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### v3.57.1 - 2018-07-27
+  * Disable ServerkeyBits on RHEL 7.4 and later
+
 ### v3.57.0 - 2017-12-10
   * Add support for AuthenticationMethods and AllowAgentForwarding
     options in sshd_config
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,6 @@
 {
   "name": "ghoneycutt-ssh",
-  "version": "3.57.0",
+  "version": "3.57.1",
   "author": "ghoneycutt",
   "summary": "Manages SSH",
   "license": "Apache-2.0",


README.md
```diff
@@ -712,6 +712,20 @@ Boolean to enable AcceptEnv options for specifying environment variables. Defaul
...
 
+sshd_acceptenv_defaults
+-------------
+Array of environment variables options for AcceptEnv setting.
+
+- *Default*: [ 'LANG', 'LC_CTYPE', 'LC_NUMERIC', 'LC_TIME', 'LC_COLLATE',
+               'LC_MONETARY', 'LC_MESSAGES', 'LC_PAPER', 'LC_NAME', 'LC_ADDRESS',
+               'LC_TELEPHONE', 'LC_MEASUREMENT', 'LC_IDENTIFICATION', 'LC_ALL' ]
+
+sshd_acceptenv_custom
```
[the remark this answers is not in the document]
renato-farias: The sshd_acceptenv option continue doing the same role. it doesnt accept string / array (ssh::sshd_acceptenv type must be true or false.)
The sshd_acceptenv_defaults is the list of defaults AcceptEnv, to keep the originality and don't brake any old installations; I think is good to keep for defaults. Imagine a huge list of servers using this module.
The sshd_acceptenv_custom option is for adding a new custom AcceptEnv Keeping the default ones.

How does it sound?

askoog:
Any plans on accepting this PR?

README.md
```diff
@@ -712,6 +712,20 @@ Boolean to enable AcceptEnv options for specifying environment variables. Defaul
...
+-------------
+Array of custom environment variables options for AcceptEnv setting. Set here a new one if you would like to keep the default ones set in sshd_acceptenv_defaults option.
```
ghoneycutt: I believe the comment should read something like

```
Array of custom environment variables options for AcceptEnv setting. If set, this will override the defaults specified by `sshd_acceptenv_defaults`.
```

manifests/init.pp
```diff
@@ -880,6 +882,17 @@
     validate_re($sshd_config_allowagentforwarding, '^(yes|no)$', "ssh::sshd_config_allowagentforwarding may be either 'yes' or 'no' and is set to <${sshd_config_allowagentforwarding}>.")
   }
 
+  $sshd_acceptenv_defaults_list = [
+    'LANG', 'LC_CTYPE', 'LC_NUMERIC', 'LC_TIME', 'LC_COLLATE',
+    'LC_MONETARY', 'LC_MESSAGES', 'LC_PAPER', 'LC_NAME', 'LC_ADDRESS',
+    'LC_TELEPHONE', 'LC_MEASUREMENT', 'LC_IDENTIFICATION', 'LC_ALL' ]
+
+  if (size($sshd_acceptenv_defaults) == 0 ) {
```
ghoneycutt: If `sshd_acceptenv_custom` is `undef` then we should use `sshd_acceptenv_defaults`, else use `sshd_acceptenv_custom`.

The logic should not assume that the defaults are always wanted and concatenate them together as that goes against the idea of specifying your own custom values if you are always stuck with the defaults.

ghoneycutt:
@askoog @renato-farias Made a couple comments that need to be addressed as well as adding spec tests, then this can be merged. Thanks for your contributions!

renato-farias:
@ghoneycutt I changed the logic using only sshd_acceptenv as option, if not set, it uses the default ones, if so, an array of envs will be applied. I fixed the fixtures too, however the CI is broken, can you check?

spec/fixtures/sshd_config_debian
```diff
@@ -99,6 +99,18 @@ UsePAM yes
 # Accept locale-related environment variables
 AcceptEnv LANG
 AcceptEnv LC_CTYPE
+AcceptEnv LC_NUMERIC
```
ghoneycutt: you shouldn't be adding to these fixtures files. These are fixtures of what works for those OS's and should not change with your PR.

templates/sshd_config.erb
```diff
@@ -154,12 +154,12 @@ GSSAPICleanupCredentials <%= @sshd_gssapicleanupcredentials_real %>
 UsePAM <%= @sshd_use_pam_real %>
 <% end -%>
 
-<% if @sshd_acceptenv_real == true -%>
+<% if @sshd_acceptenv_conf.length > 0 -%>
```
ghoneycutt: We can't change this type as it would be a breaking change. Could you just revert to what you had before and change the logic in the manifest. That would have worked.

renato-farias: I tried to avoid any braking change on my first version. As you want to keek less settings/attributes I can't see any reason to keep sshd_acceptenv as boolean and add one/two more attribute(s) for configuring it. Let's keep just one applying the defaults when it is not specified (and the OS supports, of course) and apply customized when it is.

renato-farias:
This PR was moved to https://github.com/ghoneycutt/puppet-module-ssh/pull/300

PR #291, opened 2018-09-24
Add RevokedKeys option to sshd_config

asifbadkarma:
Tested on our puppet 4 environment with rhel 6 and 7 servers.

I could test on Centos7 and Ubuntu if it helps.


asifbadkarma:
updating readme and whitespace issues

PR #300, opened 2019-01-10
module able to accept custom AcceptEnv to sshd - Fixing issue #288

renato-farias:


renato-farias:
@ghoneycutt I fixed the mess. Please take a look.

dhose:
Any plans on merging this PR? 🙂

manifests/init.pp
```diff
@@ -446,6 +447,15 @@
     }
   }
 
+  case type3x($sshd_acceptenv_list) {
```
ghoneycutt: you could just use `validate_array()`

spec/fixtures/sshd_config_debian
```diff
@@ -97,9 +97,20 @@ GSSAPICleanupCredentials yes
 UsePAM yes
 
 # Accept locale-related environment variables
-AcceptEnv LANG LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES
```
ghoneycutt: You can't change the fixtures files, they are meant to be static

manifests/init.pp
renato-farias: validate_array Deprecated: Will be removed in a future version of stdlib. See validate_legacy.

spec/fixtures/sshd_config_debian
renato-farias: I can't see any problem on it. It is according the module will generate the files.

PR #303, opened 2019-02-28
Make sure Match-configuration is at the bottom

### LINKED ISSUE #302: Match-configuration in sshd_config is not at the bottom
The Match-configuration provided by `sshd_config_match` should be at the bottom of sshd_config, but it seems that some configuration values are added beneath it. This causes sshd to crash if a Match-configuration is set in combination of e.g. HostCertificate.

peterverraedt:
Fixes #302

### DIFF
--- a/templates/sshd_config.erb
+++ b/templates/sshd_config.erb
@@ -264,15 +264,6 @@ AllowGroups <%= @sshd_config_allowgroups_real.join(' ') %>
 <% if @sshd_config_key_revocation_list_real -%>
 RevokedKeys <%= @sshd_config_key_revocation_list_real %>
 <% end -%>
-<% if @sshd_config_match -%>
-
-<%   @sshd_config_match.sort.each do |key, hash| -%>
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Thanks for the contribution! Seems testing broke. Let me fix that and then we can see about this PR.

ghoneycutt:
BTW, I'm rewriting the module entirely and in the new version you can specify Match statements and anything else at the bottom of the configs :)

ghoneycutt:
Attempting to fix spec tests with PR #304

ghoneycutt:
@peterverraedt you should look into using branches. Because you worked off of master, this would be very difficult for you to rebase, which would need to happen to pick up the fixes in PR #304.


https://guides.github.com/introduction/flow/

ghoneycutt:
Rebased in PR #305

ghoneycutt:
Released in v3.59.1

PR #307, opened 2019-03-31
V4

ghoneycutt:


spec/classes/server_spec.rb
```diff
@@ -0,0 +1,1311 @@
...
+
+  default_facts = osfamily_matrix['EL-7'].merge(defaults)
+
+  let(:facts) { default_facts }
+
+  osfamily_matrix.each do |os, facts|
+    context "with default params on osfamily #{os}" do
+      let(:facts) { defaults.merge(facts)}
+
+      # FIXME - first one fails. If you remove the duplicate, the first compile
```
ghoneycutt: @Phil-Friderici I could really use your help on this one. I can't seem to debug why the first test always fails and why passing `~` as nil/undef from hiera data in module is not working as expected.

I've been rewriting this module for Puppet 5/6 and doing it against the man pages so it supports every option and has the appropriate type checking. It also has a `custom` param so you can add whatever else to the configs.

ghoneycutt:
TODO:

manifests
- [ ] init (agent class)
- [ ] server
- [ ] config_entry (for managing `~user/.ssh/config`)

docs
- [ ] README
- [ ] CONTRIBUTING
- [ ] REFERENCE

testing
- [ ] spec tests
- [ ] acceptance testing
- [ ] vagrant

.fixtures.yml
```diff
@@ -2,18 +2,13 @@ fixtures:
...
-      repo: 'https://github.com/puppetlabs/puppetlabs-firewall.git'
-      ref: '1.9.0'
+      ref: '5.2.0'
     sshkeys:
-      repo: 'git://github.com/puppetlabs/puppetlabs-sshkeys_core.git'
-      ref: '1.0.1'
+      repo: 'https://github.com/puppetlabs/puppetlabs-sshkeys_core.git'
+      ref: '1.0.2'
+      puppet_version: '>= 6.0.0'
   symlinks:
```
Dan33l: with recent [puppetlabs_spec_helper](https://github.com/puppetlabs/puppetlabs_spec_helper/commit/b62449bc388cfc62d06630eac9f70b1b7492999e#diff-1237b5f9023710b4e56fc7899c71f98dR35) symlinks should not be necessary.

spec/classes/server_spec.rb
ghoneycutt: This seems to have to due with running puppet as a gem under osx

Phil-Friderici: @ghoneycutt After `pdk convert` I don't get that error. I only get errors for the tests of ssh_version facts. Though `pdk validate` throws alot of errors. But that should be only some minor work.

If you are ok to move to PDK (which I really recommend) you can have a look at my PR:
https://github.com/Phil-Friderici/puppet-module-ssh/pull/8

Did convert some modules to PDK already , so I am happy to help with that. Please let me know :)

Phil-Friderici: The errors for the fact disapeared after using mocha for mocking :)

Phil-Friderici: Based on the v4 branch, I have fixed the unit tests. I have created a PR with the needed changes (https://github.com/ghoneycutt/puppet-module-ssh/pull/370).

Please review or comment :)

types/yes_no.pp
```diff
@@ -0,0 +1 @@
+type Ssh::Yes_no = Enum['yes', 'no']
```
ghoneycutt, line 1: this is now in stdlib

Phil-Friderici: Thanks for the hint. But I am not sure if it would be wise to use that, it accepts yes/no in all possible cases (`Pattern[/\A(?i:(yes|no))\z/]`).

The man page for ssh_config says this:
`The possible keywords and their meanings are as follows (note that keywords are case-insensitive and arguments are case-sensitive)`

From the examples in the docs I understand that keywords are the option names and arguments are the values. In that case I wouldn't like to switch to stdlib as it would allow all cases.

ghoneycutt: Maybe we should use stdlib to match and then use a function to force the correct case.

Phil-Friderici: noooo, don't fix wrong input if we can enforce correct one in first place.
Let's follow the man page; "arguments are case sensitive".

data/os/Debian/7.yaml
```diff
@@ -0,0 +1,2 @@
+---
```
TheMeier, line 1: shouldn't that be: 
```
--- {}
```

?

Phil-Friderici:
Hej @ghoneycutt,

I have found some parameters that have been in v3 of the module but haven't added to v4 yet.

ssh_config `UseRoaming, MACs `
sshd_config `GSSAPIKeyExchange, Match, PAMAuthenticationViaKBDInt, ServerKeyBits, UsePrivilegeSeparation`

I checked man pages for them and was able to find most of them on [here](https://linux.die.net/man/5/sshd_config) but they were missing [there](https://man.openbsd.org/sshd_config).

I think we should add them to the module. They only get into the config files if the parameters are really set. So they shouldn't harm.

What is your opinion on that ?

ghoneycutt:
I'm OK with adding them. If it gets to be too much such as options only present in sun's SSH implementation you might just add those through a `custom` param.

Phil-Friderici:
@ghoneycutt we are coming to an end here :)

From my perspective the code looks very much cleaned up now. The unit test should be complete. Still missing are the documentations, would feel better if a native speaker could write them.

Regarding the default settings for different OSes: I have taken great care to take over the existing settings from the old v3 version. While adding support for Debian 11, I realised that the old settings where more like a compromise to get new OSes integrated. Now that we have everything OS related in hiera files (one for each flavour) we could change the defaults to become closer to the distribution defaults.

And one more thing: I have created a simple text file that show the ssh(d)_config directives and the parameter names for v3 and v4. I think this could be helpfull for upgrading to v4. Should I integrate it ?

Please review and comment.

ghoneycutt:
@Phil-Friderici Could you please check out my late notes - https://github.com/ghoneycutt/puppet-module-ssh/pull/373#issuecomment-1040678698 Think perhaps this can be simplified more by just removing the code around use_roaming.

Phil-Friderici:
> @Phil-Friderici Could you please check out my late notes - [#373 (comment)](https://github.com/ghoneycutt/puppet-module-ssh/pull/373#issuecomment-1040678698) Think perhaps this can be simplified more by just removing the code around use_roaming.

Removed the UseRoaming functionality in PR https://github.com/ghoneycutt/puppet-module-ssh/pull/374. Additionally I have added a bulletpoint to the TODO list above.

Phil-Friderici:
> Could you please resolve any merge conflicts so this can be merged cleanly.

From my perspective rebasing only adds a lot of extra work and brings no real benefit. v4 is a new module which is completely rewritten and not compatible with older versions anymore. I would recommend to rename the old master branch to v3 and make the v4 branch the new master.

UPGRADING.md
```diff
@@ -0,0 +1,220 @@
+# Upgrading from v3 to v4
```
ghoneycutt, line 1: This is amazing!!!

Phil-Friderici:
Hej @ghoneycutt,
what is missing here and is there something I can do to speed up the process ?
Thanks & phile grüße

Phil-Friderici:
ping

Phil-Friderici:
I am closing this PR because the v4 branch already made it into the main/master branch.

PR #309, opened 2019-04-20
Add support for debian9

hdep:
Hi,

here is a PR to add support for debian 9 which use ssh 7.4.
I try to do my best for the tests, but I'm not sure I did it right.
Please le me know ! 

Thanks


hdep:
cool I made it !
Let me know what do you think @ghoneycutt 

Regards,

ghoneycutt:
Looks good. Could you please update the `metadata.json` as well as the list of supported platforms in `README.md` and then squash your commits into one commit and I'll get this merged.

spec/fixtures/sshd_config_debian9
```diff
@@ -0,0 +1,142 @@
+# This file is being maintained by Puppet.
+# DO NOT EDIT
+
+# $OpenBSD: sshd_config,v 1.73 2005/12/06 22:38:28 reyk Exp $
```
ghoneycutt: is this file from a default, vanilla debian 9 install?

hdep: No, I jsut copy the one from debian 7 and fix the change I made on the code.
So I need to put in the original config file ?

hdep:
it doesn't compile anymore since I use default debian 9 sshd file.

spec/fixtures/sshd_config_debian9
ghoneycutt: yeah, put in the original file from a default install and add the header text to the top. The module is designed such that it uses the default values of the OS.

ghoneycutt: We also need the `ssh_config`

ghoneycutt:
You would need a conditional statement under `Debian` for Debian 9 and set any default values so that it aligns with what Debian 9 uses.

https://github.com/ghoneycutt/puppet-module-ssh/blob/master/manifests/init.pp#L195 and

ghoneycutt:
Awesome work! Reminder that we'll also need the ssh_config file

hdep:
thanks ! I added a default case statement, but this will break Debian 8 clients I'm not happy with that. To avoid this I could use the same case for Debian 7 and Debian 8.

What do you think ?

hdep:
I'm not able to configure rspec for os release major facts. Any help appreciated.

spec/classes/init_spec.rb
```diff
@@ -33,17 +33,32 @@
...
+      :sshd_service_name         => 'ssh',
+      :sshd_service_hasstatus    => true,
+      :sshd_config_fixture       => 'sshd_config_debian',
+      :ssh_config_fixture        => 'ssh_config_debian',
+    },
+    'Debian-9' => {
+      :architecture              => 'x86_64',
+      :osfamily                  => 'Debian',
+      :operatingsystemrelease    => '9',
+      :os.release.major          => '9',
```
rnelson0: I think this doesn't result in the key you want, I don’t think symbols can be dot expanded like that. You probably need the right hand side to be a hash with release as the key and a hash value with major as the key.
Alternatively, switch to rspec-puppet-facts, but that may be better served as its own separate PR.

ghoneycutt:
```ruby
:os => {
  :release => {
    major => '9',
  },
},
```

hdep: thanks guys. But it fails with this syntax.

spec/fixtures/sshd_config_debian9
```diff
@@ -0,0 +1,122 @@
...
+# Ciphers and keying
+#RekeyLimit default none
+
+# Logging
+#SyslogFacility AUTH
+#LogLevel INFO
+
+# Authentication:
+
+#LoginGraceTime 2m
```
ghoneycutt: debian9 ships with all this stuff commented out?

spec/classes/init_spec.rb
```diff
@@ -33,17 +33,40 @@
...
-      :sshd_service_name      => 'ssh',
-      :sshd_service_hasstatus => true,
-      :sshd_config_fixture    => 'sshd_config_debian',
-      :ssh_config_fixture     => 'ssh_config_debian',
+      :architecture              => 'x86_64',
+      :osfamily                  => 'Debian',
+      :operatingsystemrelease    => '7',
+      :os => {
+        :release => {
+          major => '7',
```
ghoneycutt: try putting a `:` before `major`

spec/fixtures/sshd_config_debian9
hdep: yes they explain it in the header.```
 #The strategy used for options in the default sshd_config shipped with
# OpenSSH is to specify options with their default value where
# possible, but leave them commented.  Uncommented options override the
# default value.
```

File: spec/classes/init_spec.rb (hunk 4 above)
Comment (hdep) [reply spec/classes/init_spec.rb]: done, I think this is not anymore the issue, now it fails but can't explain why as for debian 7 it is the same code, unless I did it wrong.

File: manifests/init.pp [hunk 5]
```diff
@@ -203,31 +203,45 @@
...
-      $default_sshd_pamauthenticationviakbdint = undef
-      $default_sshd_gssapicleanupcredentials   = 'yes'
-      $default_sshd_acceptenv                  = true
-      $default_service_hasstatus               = true
-      $default_sshd_config_serverkeybits       = '1024'
-      $default_sshd_addressfamily              = 'any'
-      $default_sshd_config_tcp_keepalive       = 'yes'
-      $default_sshd_config_permittunnel        = 'no'
+
+      case $::facts['os']['release']['major'] {
```
ghoneycutt: this code is against puppet3 .. use `$::operatingsystemrelease` or similar fact.

hdep: On my box this report not the valume i'm looking for : 
```
$ facter -p operatingsystemrelease
9.8
```
Instead I want to grab the major version. 
```
$ facter -p os
{
  architecture => "amd64",
  distro => {
    codename => "stretch",
    description => "Debian GNU/Linux 9.8 (stretch)",
    id => "Debian",
    release => {
      full => "9.8",
      major => "9",
      minor => "8"
    }
  },
  family => "Debian",
  hardware => "x86_64",
  name => "Debian",
  release => {
    full => "9.8",
    major => "9",
    minor => "8"
  },
  selinux => {
    enabled => false
  }
}

```
I'm running facter 5.5.13
according to https://puppet.com/docs/facter/3.11/core_facts.html#operatingsystemrelease  `$::operatingsystemrelease`  is deprectad

hdep:
Hi, If I don't use my fact it is fine.  Now I don't understand how the module works, which  value I must set or not to math my fixture file. I got this error on build     `should contain File[sshd_config] with content  supplied string (FAILED - 2)` thanks for your help

hdep:
Ok, the module set too much options by default for Debian 9 i.e `PrintMotd yes`  is kind of hardcoded so changing the behaviour seems a hughe work for me. 
Maybe the best thing is to change the debian 9 fixtures to match the default of the module. This is what is done for debian 7 I think. what do you think @ghoneycutt ?

hdep:
Well I can't make this work i'm stuck.

hdep:
Its building now. Can you please let me know what you think ? thanks !

ghoneycutt:
hdep, you need some help getting testing working on your local system?

manifests/init.pp
```diff
@@ -193,41 +193,90 @@
...
+          $default_sshd_config_xauth_location      = undef
+          $default_sshd_config_permittunnel        = undef
+          $default_sshd_config_tcp_keepalive       = undef
+          $default_ssh_package_source              = undef
+          $default_ssh_package_adminfile           = undef
+          $default_sshd_gssapikeyexchange          = undef
+          $default_sshd_pamauthenticationviakbdint = undef
+          $default_service_hasstatus               = true
+        }
+        /^[7-8].*/: {
```
ghoneycutt: we dont actually support debian 8, so this should not look like we do.

spec/classes/init_spec.rb
```diff
@@ -33,17 +33,30 @@
...
-      :operatingsystemrelease => '7',
-      :ssh_version            => 'OpenSSH_6.0p1',
-      :ssh_version_numeric    => '6.0',
-      :ssh_packages           => ['openssh-server', 'openssh-client'],
-      :sshd_config_mode       => '0600',
-      :sshd_service_name      => 'ssh',
-      :sshd_service_hasstatus => true,
-      :sshd_config_fixture    => 'sshd_config_debian',
-      :ssh_config_fixture     => 'ssh_config_debian',
+      :architecture              => 'x86_64',
```
ghoneycutt: the arrows should be aligned only in their block, ie from ``Debian-7` => {` to the closing `}`

ghoneycutt: by changing this, it adds whitespace changes that could conflict with other PR's making those harder to merge without adding anything

manifests/init.pp
hdep: Yes I understand but the module worked before with debian8 if I remove it it will breaks users, I think this a bad.

hdep: Well I learned the hard way so not sure :) Maybe for another PR

ghoneycutt:
Awesome job on adding Debian 9 support!!

manifests/init.pp
ghoneycutt: ah I see you are correct as the last logic just applied to Debian and not a specific version

ghoneycutt:
@hdep could you do the small change to the spec test regarding reverting the whitespace changes and then squash/rebase your commits into one commit and I'll get this merged.

to squash/rebase, run `git log` and find the first commit before you started. Then do `git rebase -i <commit you just looked up>`

You will see all your commits. Keep the first one and for the rest, change `pick` to `s` and save with a commit message like 'Add support for Debian 9' or similar.

hdep:
@ghoneycutt will do the rebase. I just added a fail message for default case.

ghoneycutt:
Thanks! Released in v3.60.0

PR #310, opened 2019-04-24
Add for the possibility of an IPv6 address

### LINKED ISSUE #308: Notice: /Stage[main]/Ssh/Sshkey[hostname.domain.tld]/host_aliases: host_aliases changed ['hostname', ''] to ['hostname', undef] (corrective)
If a host has no $::ipaddress, then on each and every host I get this notice on every Puppet run.

Please handle the non-existence of the $::ipaddress fact correctly. There is a good reason this host has no $::ipaddress, as this host only has IPv6 connectivity.

(For bonus points, please include the IPv6 address of hosts as an hostalias :)

bjvrielink:
Also cover the case that there is no IPv4 address on an IPv6 only host.

Fixes #308. Tested with 2 Centos7 hosts running Puppet5.5 One of them has only IPv6, the other one is dual homed. The resulting known_hosts file contains the correct amount of IP addresses.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -974,10 +974,18 @@ class ssh (
     }
   }
 
+  # If either IPv4 or IPv6 stack is not configured on the agent, the
+  # corresponding $::ipaddress(6)? fact is not present. So, we cannot assume
+  # these variables are defined. Getvar (Stdlib 4.13+, ruby 1.8.7+) handles
+  # this correctly.
+  if getvar('::ipaddress') and getvar('::ipaddress6') { $host_aliases = [$::hostname, $::ipaddress, $::ipaddress6] }
+  elsif getvar('::ipaddress6') { $host_aliases = [$::hostname, $::ipaddress6] }
+  else { $host_aliases = [$::hostname, $::ipaddress] }
+
   # export each node's ssh key
   @@sshkey { $::fqdn :
     ensure       => $ssh_key_ensure,
-    host_aliases => [$::hostname, $::ipaddress],
+    host_aliases => $host_aliases,
     type         => $ssh_key_type,
     key          => $key,
   }
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -311,6 +311,32 @@ describe 'ssh' do
       }
 
       it { should have_ssh__config_entry_resource_count(0) }
+
+      context 'with exported sshkey resources' do
+        subject { exported_resources}
+        context 'With only IPv4 address' do
+          let(:facts) { default_facts.merge( facts )}
+          it { should contain_sshkey('monkey.example.com').with(
+            'ensure' => 'present',
+            'host_aliases' => ['monkey', '127.0.0.1']
+          )}
+        end
+        context 'With dual stack IP' do
+          let(:facts) { default_facts.merge({ :ipaddress6 => 'dead:beef::1/64' }) }
+          it { should contain_sshkey('monkey.example.com').with(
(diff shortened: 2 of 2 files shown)


spec/classes/init_spec.rb
```diff
@@ -6,6 +6,7 @@
     :fqdn                   => 'monkey.example.com',
     :hostname               => 'monkey',
     :ipaddress              => '127.0.0.1',
+    :ipaddress6             => '::1',
```
ghoneycutt: please revert these additions as the tests should still pass without them.

ghoneycutt:
Could you please add a few spec tests that show what happens when you have ipv4 and v6, just v6 and just v4?

spec/classes/init_spec.rb
bjvrielink: Nope. It gives "error during compilation: Evaluation Error: Unknown variable: '::ipaddress6'".

As even Puppet 3.8.7 already knows the fact ipaddress6, I naively assumed assumed the ipaddress6 variable to be defined with respect to the changes I made in the init.pp code.

Of course, while I'm working on the spec tests I will as a default set ipaddress6 to '' for the default_facts, and provide a different value in the test itself.

bjvrielink:
> Could you please add a few spec tests that show what happens when you have ipv4 and v6, just v6 and just v4?

I'll work on it. I'm afraid writing a spec test for the exported resource is a little bit beyond my skills, so that will remain as a TODO :)

spec/classes/init_spec.rb
ghoneycutt: What if the system does not have an ipv6 address? The code should still work, which is why we shouldn't add the fact to the existing platform checks.

bjvrielink:
Was an interesting journey, but managed to add some tests. Also rebased into 1 commit.

manifests/init.pp
```diff
@@ -974,10 +974,14 @@
     }
   }
 
+  if getvar('::ipaddress') and getvar('::ipaddress6') { $host_aliases = [$::hostname, $::ipaddress, $::ipaddress6] }
```
ghoneycutt: why the use of `getvar()` ? you can do the conditionals without that.

spec/classes/init_spec.rb
```diff
@@ -311,6 +311,32 @@
       }
 
       it { should have_ssh__config_entry_resource_count(0) }
+
+      context 'with exported sshkey resources' do
```
ghoneycutt: This testing is awesome! Thanks!!

manifests/init.pp
bjvrielink: Nope. Gives 835 examples, 603 failures:
Evaluation Error: Unknown variable: '::ipaddress6'.
And on my IPv6 only system it would fail because the ::ipaddress variable is not defined. Note that the ipaddress and/or ipaddress6 facts are only present when the corresponding addresses are present. If the fact is not defined, then the variable in Puppet itself is also not defined, so you cannot simply use it in an if statement.

ghoneycutt: would you mind adding a short comment to the code with that explanation. The getvar() is not normally used and that would help anyone else reading the code.

bjvrielink: Done. Now it's just waiting for Travis to complete

ghoneycutt:
Thanks! Released in v3.60.0

Freeaqingme:
@bjvrielink @ghoneycutt Thanks for the work! I just ran into a similar error, and was happily surprised to see it was fixed only last month.

PR #311, opened 2019-04-30
Add support for Debian8

hdep:
Add support for debian 8


hdep:
Hi  @ghoneycutt here is Debian 8 support, but I don't think I can test it.

ghoneycutt:
Thank you!! Released in v3.61.0

PR #316, opened 2019-05-27
Allow 'local' and 'remote' for AllowTcpForwarding

anders-larsson:
This change allows usage of  'local' and 'remote' for the option AllowTcpForwarding


ghoneycutt:
Hi @anders-larsson , thank you for the contribution! I'm no longer accepting updates to v3 of this module as I'm getting v4 going which is feature complete with the documentation. Please take a look at PR #307 

Suggest you run this in your environment and check back on that PR.

anders-larsson:
Hi, thanks for your reply. I did not realise this was the case. I'll look into it.

Issue #242, opened 2017-07-24
Need support for X11DisplayOffset setting in sshd_config


williamkorb:
I would like to be able to add this to my Hiera data to configure the SSH daemon:

ssh::sshd_config_x11_display_offset: '30'

...which would then set this in the sshd_config:

X11DisplayOffset 30

The default is 10, but not high enough for our environment.

The changes look rather straightforward.

In init.pp:

Add parameter with default to the ssh class:

  $sshd_config_x11_display_offset         = '10',

Add handling for undef value:

  case $sshd_config_x11_display_offset {
    'unset', undef: { $sshd_config_x11_display_offset_real = undef }
    default: { $sshd_config_x11_display_offset_real = $sshd_config_x11_display_offset_real }
  }

Add validation:

  if $sshd_config_x11_display_offset_real != undef {
    validate_integer($sshd_config_x11_display_offset_real)
  }

In sshd_config.erb add this after the current commented value of X11DisplayOffset:

<% if @sshd_config_x11_display_offset_real != nil -%>
X11DisplayOffset <%= @sshd_config_x11_display_offset_real %>
<% end -%>

Thanks for this excellent module!

I'm kind of a Puppet and GitHub n00b, but I'd be happy to do a branch and submit the changes for your consideration if that's acceptable. I've never done that before, so if you have preferences as to how that is done, please let me know.

williamkorb:
Garrett,

I cloned your repo, created my own branch (named "addX11DisplayOffset") and incorporated my changes. Please let me know if I can become a contributor and submit my changes for your approval.

Thanks,
Bill

williamkorb:
Garrett,

I have this change coded. Can you please grant me contributor status so that I can create a new pull request to submit my changes for review?

Thanks,
Bill

ghoneycutt:
Anyone can submit a PR. Please do! It will need corresponding spec tests.

williamkorb:
OK, I created a PR. Can you point me at something that will tell me how to create the spec test?

ghoneycutt:
Check out this commit - https://github.com/ghoneycutt/puppet-module-ssh/commit/1c0d3f6bdb87aa26c21379a5230d1e10bb8f47ca

Phil-Friderici:
Fixed in v4 branch

PR #317, opened 2019-05-30
Add support for x11 display offset setting in sshd.config.

williamkorb:
Garrett,

Sorry it took me so long to get back to this. I have fixed the typo you discovered and added spec tests to address issue #242.

BTW, it had been so long since I'd worked on it that I deleted my initial repo and re-forked it as of today so as to keep things simpler. Not sure that was the correct approach since it invalidates my previous PR...

Please advise if further work is needed.

Thanks,
Bill

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,6 +441,14 @@ This module sets this option to 'yes' on Linux and undef on Solaris.
 
 - *Default*: 'USE_DEFAULTS'
 
+sshd_config_x11_display_offset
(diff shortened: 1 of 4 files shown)


williamkorb:
Munged the spec test...fixing that now...sorry for jumping the gun.

williamkorb:
Submitting a new one after fixing spec test.

PR #318, opened 2019-05-30
Add support for X11DisplayOffset setting in sshd_config

williamkorb:
Garrett,

OK, I have addressed the issue you previously pointed out and added tests which are now all passing.

This should resolve issue #242.

Please let me know if you need anything else from me on this.

Thanks,
Bill


Phil-Friderici:
v4 does allow to set `X11DisplayOffset` in sshd_config with the `$ssh::server::x11_display_offset` parameter.

Issue #233, opened 2017-06-27
support multiple sshd daemon


khaefeli:
Hi,

first: thank you for this nice module! 
we're using it on many nodes.

is it planed to support multiple ssh daemons in future?
I would like to configure >=2 ssh daemons on one node for ssh fileexchange (including chrooted directories)

cheers,
Kevin

ghoneycutt:
I don't quite follow you. Can you give an example of how the configs would change?

khaefeli:
we have multiple ` ssh_config_xy`  in our ` /etc/ssh` 

running process:
` /usr/sbin/sshd -f /etc/ssh/sshd_config_xy-o PidFile=/var/run/sshd-xy.pid`

differences: 

- ChrootDirectory
- AuthorizedKeysFile
- Port

ghoneycutt:
We would need to change quite a bit. Thinking the configs would need to be defined types, so you could call them multiple times. Is it multiple ssh_config and sshd_config? Curious what your use case is that you run SSH like this. Could you elaborate?

khaefeli:
sorry, that's a typo. I mean sshd_config_xy.

we use it as a "lightweight" file exchange with keyauth / chroot. (SFTP)
actually some ftp servers have the same functionality

joernott:
As I am in the same situation, I have forked the module and started splitting things up. In [my fork https://github.com/joernott/puppet-module-ssh/tree/multiple_sshd_configs](https://github.com/joernott/puppet-module-ssh/tree/multiple_sshd_configs), I have moved the sshd_config handling to a defined type which is called by init. The same applies for service. This however needs some more expansion, as I want to run different services with different configs in the future. Also, package has been moved to a separate class, as someone might want to install the software and only wants to configure the sshd_config and service without managing the ssh_config.

ghoneycutt:
Closing due to age. The module has undergone a major release and now supports every documented configuration parameter for OpenSSH. The module also allows for custom entries to future proof functionality and support other SSH implementations. See [UPGRADING.md](https://github.com/ghoneycutt/puppet-module-ssh/blob/main/UPGRADING.md).

PR #322, opened 2019-07-22
Multiple sshd configs, solves #233

joernott:
This pull request implements a solution to issue #233. As we also need to handle one or more additional instances of sshd as sftpserver, I split the monolithic init.pp into several pieces while preserving backwards compatibility. 

- package.pp manages the package installation, it is called by init.pp and required by the new instance.pp
- service.pp is now a defined type handling the service, its used by inid and instance
- sshd_config.pp is now a defined type and used by init and instance
- service_instance.pp creates a new systemd service (as I don't have access to Solaris, I can't add support for something comparable, however, the code is prepared for it)
- instance.pp is a wrapper to create a new sshd_config (sshd_config.$title), a service_instance and manages the service for that instance.


ghoneycutt:
The vast majority of use cases is to not run multiple instances so I think this would really add complexity. The module already lets you write multiple config files so you could implement service management at the profile level for the multiple instances.

PR #324, opened 2019-07-31
Unset default_sshd_config_serverkeybits for Ubuntu 18.04

diranged:
Ubuntu 18.04's version of SSH deprecated the ServerKeyBits setting - so we should not be setting it.


cervigni:
Hello,
Could you please merge this as important for us as well?

Thanks

Issue #314, opened 2019-05-15
Not working with Ubuntu 14.04 


flaporta:
init.pp missing code to support Ubuntu14.04. The module states that 14.04 is compatible.

witchbutter:
I had this same issue, you have to revert to 3.59.0 in order to continue working with 14.04

efungsh:
Regression appears to have occurred when Debian 9 support was added: dce102f33ad87da6151a4d7108fd36f01c5a8e63

Default behaviour for Debian was changed from a generic set of values to a fail scenario.

Looks like following supported Ubuntu versions are affected
* Precise (12.04)
* Lucid (14.04)

efungsh:
PR #325 lodged to resolve this

TheMeier:
If I understand the infos on https://wiki.ubuntu.com/Releases correctly 12.04 is EOL and even Extended Security Maintenance ended in 2019

efungsh:
Yeah, that's right.

However, this module lists it as officially supported, so the removal was inadvertent.
Restoring it was just adding back configuration defaults.

I think making a call to remove support for 12.04 LTS should be planned and it should be announced in advance.

Phil-Friderici:
Support for Ubuntu 14.04 and 16.04 was removed.
v4 supports Ubuntu 18.04 to 20.04.
PR https://github.com/ghoneycutt/puppet-module-ssh/pull/398 will add support for 22.04.

PR #325, opened 2019-08-06
#314 Restore defaults for Ubuntu 12.04 and 14.04

efungsh:
Added an $operatingsystemrelease case to restore Ubuntu 12.04 and 14.04 compatitibility.
Values based off defaults originally in 9832dee0a07d82e6f2f215fce10149fbb24aa573

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -205,7 +205,7 @@ class ssh (
       $default_service_name                    = 'ssh'
 
       case $::operatingsystemrelease {
-        /^1[24].04$/: {
+        /^1[24]\.04$/: {
           $default_sshd_config_hostkey                = [ '/etc/ssh/ssh_host_rsa_key' ]
           $default_ssh_config_hash_known_hosts        = 'no'
           $default_sshd_config_xauth_location         = '/usr/bin/xauth'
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -199,6 +199,28 @@
       $default_service_name                    = 'ssh'
 
       case $::operatingsystemrelease {
+        /^1[24].04$/: {
```
anders-larsson, line 202: You might want to escape the dot to ensure it is an actual dot when evaluated.

anders-larsson:
Hello. We've also been hit by this subtly removal of compatibility.  Looks good to me other the nitpicking added in comment :)

manifests/init.pp
```diff
@@ -205,6 +205,28 @@
       $default_service_name                    = 'ssh'
 
       case $::operatingsystemrelease {
+        /^1[24]\.04$/: {
```
ghoneycutt, line 208: Instead of a regex, a list would be easier to read.

```puppet
'12.04', '14.04': {
```

ghoneycutt (changes requested):

ghoneycutt:
I'm not interested in supporting these EOL platforms.

PR #327, opened 2019-09-24
Made host_aliases configurable

Fabian1976:
We use IPA for our host registration and authentication. When the SSH module adds the FQDN and the HOSTNAME (without domain suffix) to the global known_hosts, auto-complate with SSH only uses the HOSTNAME.
We want to be able to remove HOSTNAME and only have FQDN and IPADDRESS in the know_host_file.
With this addition the default behaviour stays the same but we have the option to alter the host_alias.

Could you please allow this change?

I didn't change the indentation of the parameters on purpose so that it is easy to see what has changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,6 +637,16 @@ forwarding is permitted. Valid values are 'yes' and 'no'.
 
 - *Default*: undef
 
+host_aliases
+------------
+Add custom host aliases. When none is provided, the host_alias will be:
+```
+FQDN,SHORT_HOSTNAME,IPADDRESS
+```
+
+- *Default*: undef
+
+
 config_entries
 --------------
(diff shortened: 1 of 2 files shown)


Phil-Friderici:
v4 does not support for exporting and importing SSH keys anymore.

Fabian1976:
Ok, so when i write a new PR for version 4, are you going to wait another 3+ years and then "suddenly" react and decline because it was written for an old version?

Or is there another reason why it was not merged in the first place?

Phil-Friderici:
The functionality to export and import SSH keys was causing a lot of issues in large environments with constantly changing servers. Therefore it was removed in v4.

I can not speak for the module maintainer (@ghoneycutt) but I think it should be possible to re-add the functionality when it is disabled by default.

Issue #320, opened 2019-07-15
Add SLES 15 support


absld:
Hi, I tried your module on SLES 15 and it mostly works fine. The only thing I noticed was that you should change parameter $default_sshd_config_subsystem_sftp in manifests/init.pp to '/usr/lib/ssh/sftp-server' as you already did for SLES 12.

PR #331, opened 2020-01-22
Added sles15 resolved issue #320

FcoCalero:
Resolved Deprecated option ServerKeyBits for sles15
Added #320


manifests/init.pp
```diff
@@ -171,17 +171,27 @@
...
+            case $::operatingsystemrelease {
+              /12\./: {
+                $default_sshd_config_subsystem_sftp = '/usr/lib/ssh/sftp-server'
+                $default_sshd_config_serverkeybits  = '1024'
+              }
+              /15\./: {
+                $default_sshd_config_subsystem_sftp = '/usr/lib/ssh/sftp-server'
+                $default_sshd_config_serverkeybits  = undef
+              }
+              default: {
```
ghoneycutt, line 190: The default is the same as `12` so suggest using `default` and `15`.

ghoneycutt:
Hi @FcoCalero Thanks for the contribution. This will also need tests added as well as rebased against master to pick up fixes for TravisCI.

Check out https://github.com/ghoneycutt/puppet-module-ssh/blob/master/spec/classes/init_spec.rb#L196

FcoCalero:
Yeah sorry, i clicked it by accident and i didn't see the way to cancel it,
i will do something probably by next week or so.

manifests/init.pp
```diff
@@ -176,18 +176,24 @@
...
       $default_sshd_config_include             = undef
       case $::architecture {
         'x86_64': {
-          if ($::operatingsystem == 'SLES') and ($::operatingsystemrelease =~ /^12\./) {
-            $default_sshd_config_subsystem_sftp = '/usr/lib/ssh/sftp-server'
-          } else {
-            $default_sshd_config_subsystem_sftp = '/usr/lib64/ssh/sftp-server'
+          if ($::operatingsystem == 'SLES') {
+            case $::operatingsystemrelease {
+              /15\./: {
```
anders-larsson, line 188: Is it intentional to break SFTP support for SLE[DS] 12? It uses the wrong path after this change. Currently SLES12 is still stated as supported in metadata.json.

Sorry for writing in an old pull request though.

PR #335, opened 2020-02-27
Explicitly unset the ipaddress6 fact

vStone:
Not sure when this behavior changed...


vStone:
https://github.com/rodjek/rspec-puppet/compare/v2.7.7...v2.7.8

PR #340, opened 2020-06-03
Add support for Debian 10 buster

bahner:
SSIA Added Debian 10 with the actual settings and mode from first install


bahner:
I fail to understand the error in Travis. Seems to fail on Xenial, but that is hardly relevant.

ghoneycutt:
Thanks for the contribution! This work has already been done. You can follow along with #341

PR #348, opened 2020-10-13
add option to enable subsystem sftp in sshd config

eylenth:
It could be handy to have an option to disable or enable the sftp Subsystem in the sshd config . 

https://serverfault.com/questions/653812/enable-ssh-shell-access-but-disable-sftp-access


fnollet:
Can we have some feedback here to merge this?

Phil-Friderici:
v4 does allow to unset `Subsystem` in sshd_config by unsetting the `$ssh::server::subsystem` parameter.

Issue #330, opened 2019-12-12
Add support for HostKeyAlgorithms or config options not in template


pynixadm:
Hey,

I've been puzzling over this all day, the only option missing is a way to restrict HostKeyAlgorithms similar to ciphers/kex/macs.

I need to be able to add this line:

```
HostKeyAlgorithms ssh-ed25519-cert-v01@openssh.com,ssh-rsa-cert-v01@openssh.com,ssh-ed25519,ssh-rsa
```

I'm not very proficient with git, but I could look into the code tomorrow and see if I can provide something.

Any help would be greatly appreciated!

Cheers.

pynixadm:
Here is what I came up with:

**init.pp**
```
  $ssh_config_hostkeyalgorithms          = undef,
  $sshd_config_hostkeyalgorithms         = undef,

  [...]
  
  if $ssh_config_hostkeyalgorithms != undef {
    validate_array($ssh_config_hostkeyalgorithms)
  }

  if $sshd_config_hostkeyalgorithms != undef {
    validate_array($sshd_config_hostkeyalgorithms)
  }
```

**ssh_config.erb**
```
<% if @ssh_config_hostkeyalgorithms -%>
  HostKeyAlgorithms <%= @ssh_config_hostkeyalgorithms.join(',') %>
<% end -%>
```

**sshd_config.erb**
```
<% if @sshd_config_hostkeyalgorithms -%>
HostKeyAlgorithms <%= @sshd_config_hostkeyalgorithms.join(',') %>
<% end -%>
```

**README.md**
```
ssh_config_hostkeyalgorithms
---------------
Array of hostkey algorithms to be used with the HostKeyAlgorithms option in ssh_config.

- *Default*: undef

[...]

sshd_config_hostkeyalgorithms
---------------
Array of hostkey algorithms to be used with the HostKeyAlgorithms option in sshd_config.

- *Default*: undef

```

I've tested this in my testlab puppet 5.5 master and it passes.

Cheers!

16c7x:
@pynixadm  Have you put a pull request in for this change?

Phil-Friderici:
Functionality to set `HostKeyAlgorithms` via `$ssh::server::host_key_algorithms` was added with v4.
Please re-open if this doesn't fit your use case.

PR #352, opened 2021-07-02
Issue #330 add support for HostKeyAlgorithms

16c7x:
Add support for HostKeyAlgorithms or config options not in template #330
@pynixadm has proposed a solution that works but didn't create a pull request for it so I have.

### DIFF
--- a/README.md
+++ b/README.md
@@ -639,6 +639,12 @@ forwarding is permitted. Valid values are 'yes' and 'no'.
 
 - *Default*: undef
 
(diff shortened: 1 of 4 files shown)


Phil-Friderici:
Closing this PR as it was made for the v3 version of the module which is outdated now. v4 does allow to set `HostKeyAlgorithms` in sshd_config with the `$ssh::server::host_key_algorithms` parameter.

PR #357, opened 2021-09-22
A number of changes

jovandeginste:
Hi,

We accrued a number of small changes over the past years, which we failed to provide upstream PR's for. Would you prefer smaller PR's, or is this single PR acceptable?


jovandeginste:
I split the changes into smaller PR's

PR #364, opened 2021-11-02
Update init.pp

jonathan-dill-nih:
serverkeybits is deprecated in 7.4 but versioncmp($::operatingsystemrelease, '7.4') < 0 is false for EL 7.4 should be <= 0

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -147,7 +147,7 @@ class ssh (
       $default_sshd_gssapicleanupcredentials   = 'yes'
       $default_sshd_acceptenv                  = true
       $default_service_hasstatus               = true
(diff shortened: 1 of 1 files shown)


jonathan-dill-nih:
A more universal fix might be versioncmp($::ssh_version_numeric, '7.6') < 0 || "${::ssh_version}" =~ /^OpenSSH_7.6p/ all the different numbering schemes for OS versions get a bit messy, need to check my syntax. It's unfortunate that there is no way to distinguish the p1 via $::ssh_version and plain 7.6 does not have it.

jonathan-dill-nih:
Nevermind I was reading the rule as the inverse.

PR #365, opened 2021-11-02
ServerKeyBits without $::operatingsystemrelease

jonathan-dill-nih:
Attempt to fix ServerKeyBits for all OS with OpenSSH, sparing Sun_SSH which might use ServerKeyBits 768, although sensible folks should have already disabled SSHv1. In theory, this should work for both EL and Debian derivatives regardless of the OS version numbering scheme. Covering possibly mythical OpenSSH_7.6 without p1.


Phil-Friderici:
Closing this PR as it was written for the (now outdated) old version 3 of this moudle.
v4 of this module does configure Solaris by default to use `ServerKeyBits 768`.

PR #367, opened 2021-12-27
Allow not exporting SSH keys

mattock:
This enables silencing some useless warnings when using Puppet Bolt for example

Signed-off-by: Samuli Seppänen <samuli.seppanen@puppeteers.net>


ghoneycutt:
Thank you!

PR #368, opened 2022-01-09
Remove sort in sshd_config_match as the orders matters

philippdieter:


Phil-Friderici:
Closing this PR as it was writen for the old v3 version of this module and I think the functionality can be reached with the new `$custom` parameter as well.
If this doesn't fit your needs, please create a new PR against the new version.

PR #369, opened 2022-01-10
Restore SFTP functionality for SLE12

anders-larsson:
This restores SFTP subsystem path for SLE12.

Changed the logic to use versioncmp so all future versions of SLE will continue to use the new location (until it changes again...)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -184,15 +184,15 @@ class ssh (
       case $::architecture {
         'x86_64': {
           if ($::operatingsystem == 'SLES') {
-            case $::operatingsystemrelease {
-              /15\./: {
-                $default_sshd_config_subsystem_sftp = '/usr/lib/ssh/sftp-server'
-                $default_sshd_config_serverkeybits  = undef
-              }
-              default: {
-                $default_sshd_config_subsystem_sftp = '/usr/lib64/ssh/sftp-server'
-                $default_sshd_config_serverkeybits  = '1024'
-              }
+            if versioncmp($::operatingsystemrelease, '15') >= 0 {
+              $default_sshd_config_subsystem_sftp = '/usr/lib/ssh/sftp-server'
+              $default_sshd_config_serverkeybits  = undef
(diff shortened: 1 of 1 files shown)


anders-larsson:
Created because it was broken in #331  however it looks like SLES12 is no longer supported[1] by its vendor.

[1]: https://www.suse.com/lifecycle/

anders-larsson:
Ugh. Now it should be back to where it was before and including #331. server_bits is set to undef on SLES15 and newer while set to 1024 on SLES12 and older.

ghoneycutt:
Hi @anders-larsson this module is going through some big changes by our friend @Phil-Friderici so I will defer to him when this should get merged.

Phil-Friderici:
@anders-larsson These details have been moved into hiera in the v4 branch \o/ The values were transfered from the latest v3 branch.

When I do understand you correct that means that SLED/SLES 12 also needs
`/usr/lib/ssh/sftp-server` instead of
`/usr/lib64/ssh/sftp-server` ?

Please see the moved data [here](https://github.com/ghoneycutt/puppet-module-ssh/tree/v4/data/os/)

anders-larsson:
> @anders-larsson These details have been moved into hiera in the v4 branch \o/ The values were transfered from the latest v3 branch.
> 
> When I do understand you correct that means that SLED/SLES 12 also needs `/usr/lib/ssh/sftp-server` instead of `/usr/lib64/ssh/sftp-server` ?
> 
> Please see the moved data [here](https://github.com/ghoneycutt/puppet-module-ssh/tree/v4/data/os/)

That's correct. Everything 12 and newer should use `/usr/lib/ssh/sftp-server`

Phil-Friderici:
Fixed in v4 with https://github.com/ghoneycutt/puppet-module-ssh/pull/373/commits/44a5ba7871f1a146e721ee1fee872fed1fc23188

Phil-Friderici:
I don't think there will be another v3 release. So if you are ok to have this fix in the upcoming v4 release you can close this PR :)

PR #371, opened 2022-01-18
Refactor unit tests and minor fixes

Phil-Friderici:


metadata.json
```diff
@@ -85,8 +85,6 @@
     {
       "operatingsystem": "Ubuntu",
       "operatingsystemrelease": [
-        "12.04",
```
ghoneycutt, line 88: These old platforms are only here for you, so feel free to remove anything that is end of life that you are no longer supporting.

Phil-Friderici: Thanks, good to know :)

ghoneycutt (approved):

spec/classes/init_facterdb_spec.rb
```diff
@@ -0,0 +1,395 @@
...
+            'owner'   => 'root',
+            'group'   => 'root',
+            'mode'    => '0644',
+            'require' => 'Package[openssh-clients]',
+          },
+        )
+      end
+
+      # FIXME: add test for
+      # Sshkey <<||>> {
```
ghoneycutt, line 99: We should just leave this collection functionality out and document in the readme and let people implement in a profile.

Phil-Friderici: Do you mean that we should remove the functionality from the manifest ?
https://github.com/ghoneycutt/puppet-module-ssh/blob/v4/manifests/init.pp#L670-L675

Doesn't it makes sense to completly remove both functionalities ($ssh_key_import & $key_export) then ?

ghoneycutt: Yep! There have always been issues around these and just showing people how to do it would be better.

Phil-Friderici: added a check box for adding the documentation in the main PR https://github.com/ghoneycutt/puppet-module-ssh/pull/307

Will remove both functionalities eventualy.

metadata.json
Phil-Friderici: @anders-larsson please let me know which OS versions do you would like to get support for.

anders-larsson: *From our perspective on what can be removed*
* Solaris (not used at all. Maybe someone else does?)
* SLE[DS]10 (all other releases supported[1], SLED11 is not supported but SLES11 is with LTSS)
* Debian (probably same as Ubuntu though, keep it)
* Ubuntu 12.04 (looks like all other versions are still supported[2])

[1]: https://www.suse.com/c/explore-options-for-suse-linux-enterprise-11-service-pack-4-end-of-life-mar-31-2019/
[2]: https://ubuntu.com/about/release-cycle

.fixtures.yml
```diff
@@ -2,11 +2,11 @@ fixtures:
   repositories:
     stdlib:
       repo: 'https://github.com/puppetlabs/puppetlabs-stdlib.git'
-      ref: '5.2.0'
+      ref: 'v7.1.0'
```
anders-larsson, line 5: Are we using some functionality that requires more recent versions of stdlib (and other fixtures) or are this just a version bump?

Phil-Friderici: This is just a version bump.
If we want to use stdlib::Yes_no we need v6.1.0. Otherwise we should be good with 5.2.0.

metadata.json
Phil-Friderici: Thanks Anders ! Will re-add Ubuntu 14.04 later on.

Phil-Friderici: If Ubuntu 14.04 needs any specific tweaks we need to add them later. There isn't any specific handling in the code. Guessing that the default values are valid for it.

anders-larsson: > If Ubuntu 14.04 needs any specific tweaks we need to add them later. There isn't any specific handling in the code. Guessing that the default values are valid for it.

This could be useful https://github.com/kodguru/puppet-module-ssh/commit/1159fed26260babc7c9e663b0d092dfd9b9be16b

Phil-Friderici: Great and perfect timing. Just realised that the old unit tests doesn't verify Ubuntu 12 nor 14 and the code actually fails to run on them.

Are these values from a stock Ubuntu 14 system ? Or is it tweaked in any way ?

anders-larsson: I'm not 100% sure. I just re-added the old configuration that was previously present in this module.

Phil-Friderici:
The unit tests should now be good and working. I will merge these commits into the v4 branch and start a new PR for the next steps.

PR #372, opened 2022-01-27
Moving data from manifests to hiera and minor fixes

Phil-Friderici:
Here I will collect the commits needed to move the data from manifests to hiera.


.sync.yml
```diff
@@ -17,7 +17,7 @@ appveyor.yml:
 
 spec/spec_helper.rb:
   coverage_report: true
-  minimum_code_coverage_percentage: 65
+  minimum_code_coverage_percentage: 57
```
ghoneycutt, line 20: This seems oddly specific :>

Phil-Friderici: forgive me ;)

PR #373, opened 2022-02-09
Further working on v4

Phil-Friderici:
Use this PR to collect further commits to get v4 done.


Phil-Friderici:
@ghoneycutt The $use_roaming parameter should only be used when OpenSSH 5.4 (or later is used). Currently there  is this code block taking care for it:
```
  # UseRoaming should only be used if OpenSSH 5.4 is used on the client
  if $use_roaming != undef {
    $use_roaming_real = $use_roaming
  } else {
    if $::ssh_version =~ /^OpenSSH/ and versioncmp($::ssh_version_numeric, '5.3') == 1 {
      $use_roaming_real = 'no'
    } else {
      $use_roaming_real = undef
    }
  }
```
This is the last parameter not handled via hiera. It should be possible to move that logic into hiera by adding another level to hiera. Additionally I think it would be needed to add a fact to specify the used SSH flavour (eg: OpenSSH, Sun SSH) and facts for SSH major and minor versions.

What would you recommend ?

Phil-Friderici:
In the meantime I am getting the feeling we should remove the code block for `UseRoaming`. In it current state it adds the directive automatically if OpenSSH 5.4 is used, ignoring the defaults used from the OSes.

ghoneycutt:
> @ghoneycutt The $use_roaming parameter should only be used when OpenSSH 5.4 (or later is used). Currently there is this code block taking care of it:
> 
> ```
>   # UseRoaming should only be used if OpenSSH 5.4 is used on the client
>   if $use_roaming != undef {
>     $use_roaming_real = $use_roaming
>   } else {
>     if $::ssh_version =~ /^OpenSSH/ and versioncmp($::ssh_version_numeric, '5.3') == 1 {
>       $use_roaming_real = 'no'
>     } else {
>       $use_roaming_real = undef
>     }
>   }
> ```
> 
> This is the last parameter not handled via hiera. It should be possible to move that logic into hiera by adding another level to hiera. Additionally I think it would be needed to add a fact to specify the used SSH flavour (eg: OpenSSH, Sun SSH) and facts for SSH major and minor versions.
> 
> What would you recommend ?

Recommend that we document the need for the specific version instead of using code as the parameter to control this will be set depending on the platform. No need to determine this with facts and have another layer in hiera to set it.

PR #375, opened 2022-02-23
Prepare the docs

Phil-Friderici:


README.md
```diff
@@ -2,43 +2,43 @@
 
 Manage ssh client and server.
 
-The module uses exported resources to manage ssh keys and removes ssh keys that
-are not managed by puppet. This behavior is managed by the parameters
-ssh_key_ensure and purge_keys.
+This module is based on the OpenSSH v7.0 implementation. All parameters that are described in the
+man pages are available to this module with the exception of the Match parameter. Some SSH
+implementations does provide extra features and use additional parameters. These deviations can
+still be managed with the help of the $custom parameter. This freetext parameter allows you to add
```
ghoneycutt, line 8: add backticks to make `$custom`

README.md
```diff
@@ -49,831 +49,17 @@ Unsupported parameters for ssh_config:
 AddressFamily, Tunnel, TunnelDevice, PermitLocalCommand, HashKnownHosts
 
 Unsupported parameters for sshd_config:
-KerberosOrLocalPasswd, KerberosTicketCleanup, KerberosGetAFSToken, TCPKeepAlive, ShowPatchLevel, MaxSessions, PermitTunnel
+KerberosOrLocalPasswd, KerberosTicketCleanup, KerberosGetAFSToken, TCPKeepAlive, ShowPatchLevel,
```
ghoneycutt, line 52: Are these unsupported really as we have the custom param? Perhaps this area can be removed.

ghoneycutt:
This is awesome! It should make it much easier for the community to add platforms.

README.md
Phil-Friderici: I agree it, this piece of info seems a little bit lonely here, will remove it. Sure the provided default settings doesn't use them.

Also I have added that this module is based around OpenSSH (but can be used for any other as well).

Phil-Friderici:
Glad you like it. Easy maintenance is one of the key intentions here.

I have an idea and would like to know what you think of it:
We have removed all these nasty checks for compatibility of SSH parameters. That makes the code much easier to read and maintain. On the other side there is a risk of a misconfigured SSH daemon not re-starting successfully which then renders systems to become unreachable for manual fixing.
So I am wondering myself if we could add a passive class with these sanity checks. People who like the checks would need to activate that extra class.
I guess it shouldn't be too much effort and it should only cover the real killer combinations.

README.md
Phil-Friderici: did you see that this is a list of incompatible parameters that should be used with Sun SSH implementations.

ghoneycutt: I did not.. a `### SunSSH` subsection would be helpful there.

Phil-Friderici:
> Guessing this would be things like `if $ssh_version < X, then param Y must be undef` ? I'm OK with a class that does all that so long as you can disable its inclusion as over time or the addition of platforms it could get hard to maintain.

I am not sure if these sanity checks where one of the 'selling points' of this module. Maybe we should wait if the users  do request the sanity checks before adding something that isn't asked for. Also I would like to get this work to an end soon :)

README.md
```diff
@@ -2,880 +2,78 @@
...
-manage_service
---------------
-Manage the sshd service through this module or not.  Valid values are 'true' and 'false'.
-
-- *Default*: 'true'
-
-===
 # Manage user's ssh_authorized_keys
-This works by passing the ssh::keys hash to the ssh_authorized_keys type with create_resources(). Because of this, you may specify any valid parameter for ssh_authorized_key. See the [Type Reference](http://docs.puppetlabs.com/references/stable/type.html#ssh_authorized_key) for a complete list.
+This works by passing the ssh::keys hash to the ssh_authorized_keys type with create_resources().
```
ghoneycutt, line 73: We dont actually use the old `create_resources()` anymore thanks to you changing it to native iteration.

Phil-Friderici: 👍 good point, also fixed the URL

PR #379, opened 2022-05-26
Add support for Debian 11 and Arch Linux

galenguyer:


Phil-Friderici:
Closing this PR as it was made for the v3 version of the module which is outdated now. I have added support for Archlinux, Debian 11, and Ubuntu 22.04 in https://github.com/ghoneycutt/puppet-module-ssh/pull/398.

PR #385, opened 2022-08-16
Fix Subsystem sftp and serverkeybits on SLES 12

anders-larsson:
This pull request restores path to subsystem sftp binary on SLES 12. It also restores the previous configuration where ServerKeyBits are set to 1024.

```
# cat /etc/os-release 
NAME="SLES"
VERSION="12-SP5"
...
# ls -l /usr/lib64/ssh/sftp-server
ls: cannot access '/usr/lib64/ssh/sftp-server': No such file or directory
# ls -l /usr/lib/ssh/sftp-server
-rwxr-xr-x 1 root root 445712 Nov 26  2021 /usr/lib/ssh/sftp-server
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -186,9 +186,12 @@ class ssh (
         'x86_64': {
           if ($::operatingsystem == 'SLES') {
             case $::operatingsystemrelease {
-              /15\./: {
+              /^(12|15)\./: {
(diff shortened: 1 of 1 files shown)


anders-larsson:
Closing as this seems to have been fixed by v4 release.

PR #390, opened 2022-11-05
V4 again take 2

treydock:


Phil-Friderici:
I am very happy to see progress with v4, but it feels a bit sad that all the work I had put in it is now lost because of the rebase.

treydock:
@Phil-Friderici I modified the commit to make you author and added myself as co-author.

canihavethisone:
Why is puppet version restricted to 7 in this pr? Is it no longer compatible with v6?

treydock:
I can't speak to why this was restricted to Puppet 7 as I just took over the v4 refactoring and only added RHEL8 bits.  Generally speaking if a module supports Puppet 7, it will also work with Puppet 6 as there were not significant changes between those Puppet versions that would cause modules to break.  In the context of this module, it's just about what is tested against.

canihavethisone:
@treydock @ghoneycutt I'm concerned about the **minimum** stdlib version being 8. Is this absolutely required? Also if its compatible with puppet 6 it would be good to have that in the metadata

ghoneycutt:
@canihavethisone Puppet 6 is deprecated and we are not spending any effort to support it. As Trey mentioned, it will likely "just work" though you may have to adjust your module management processes to allow for it and your other older dependencies such as stdlib.

canihavethisone:
@ghoneycutt according to https://puppet.com/docs/puppet/6/platform_lifecycle.html puppet 6 is in support until July 2023.

The bigger concern to me is having the lowest version of stdlib at v8.  If this is not required for new functions, can you please consider lowering it to v6 to ensure backwards compatibility?

PR #391, opened 2022-12-03
ssh::server::host_certificate should be multi-value

kevinsumner:
Parameter documentation implies that it should have been multi-value to begin with, and sshd_config accepts multiple HostCertificate directives, usually to handle different key types (e.g., RSA, ECDSA, and ED25519).

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -499,7 +499,7 @@ class ssh::server (
   Optional[Array[String[1]]] $hostbased_accepted_algorithms = undef,
   Optional[Ssh::Yes_no] $hostbased_authentication = undef,
   Optional[Ssh::Yes_no] $hostbased_uses_name_from_packet_only = undef,
-  Optional[String[1]] $host_certificate = undef,
+  Optional[Array[String[1]]] $host_certificate = undef,
   Optional[Array[String[1]]] $host_key = undef,
   Optional[String[1]] $host_key_agent = undef,
   Optional[Array[String[1]]] $host_key_algorithms = undef,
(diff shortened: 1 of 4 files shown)


kevinsumner:
I've treated this as an API bug, but let me know if this needs updating to maintain backward-compatibility.

Phil-Friderici:
Shouldn't `HostKey` be used to specify multiple keys instead ?
Wasn't able to find examples that are using multiple `HostCertificate` directives.

kevinsumner:
Using multiple key types (`rsa`, `ecdsa`, and `ed25519` is the default) requires multiple host keys. As I understand it, a host certificate can only be associated with a single key, so multiple certificates are needed to provide for the different key types. So, effectively, both `HostKey` and `HostCertificate` directives need to be specified multiple times when using multiple key types. The default value for `HostKey` is multi-valued, one for each key type, so this rarely has to be overridden.

I have sshd_config deployed to use host certs on a few of my machines for testing, using the branch this PR originates from. I wrote another Puppet module to use an SSH CA key, stored in eyaml, to generate host certificates. As a practical example, here's what the `HostCertificate` directives from /etc/ssh/sshd_config look like:

```
root@XXX:~# grep HostCertificate /etc/ssh/sshd_config
HostCertificate /etc/ssh/ssh_host_rsa_key-cert.pub
HostCertificate /etc/ssh/ssh_host_ecdsa_key-cert.pub
HostCertificate /etc/ssh/ssh_host_ed25519_key-cert.pub
```

Phil-Friderici:
Created #405 which will create one HostCertificate line for each file.

PR #394, opened 2023-01-03
Preparation needed to add new OS flavours (ready to merge)

Phil-Friderici:
This change add/changes those:
* Migrate from os.family to os.name
  Make sure that we can have specific configurations for each RedHat based OS.

* Add possibility to manage configuration files in .d directories
  Modern versions of SSH allow to use configuration files via include directive. Several OS vendors already use this feature.

* Move default value for $host into hiera data
  Allow host directive to be unset in ssh_config. This is needed for OS flavours that do not set this directive in ssh_config but in .d directories.

* Add OracleLinux and Scientific Linux to metadata

This change is a preparation to include support for RedHat 9, Rocky, AlmaLinux in https://github.com/ghoneycutt/puppet-module-ssh/pull/398.


Phil-Friderici:
Closed as duplicate of https://github.com/ghoneycutt/puppet-module-ssh/pull/398

PR #396, opened 2023-01-03
Add support for Rocky and AlmaLinux

Phil-Friderici:
based on https://github.com/ghoneycutt/puppet-module-ssh/pull/394

### DIFF
--- a/data/os/AlmaLinux/8.yaml
+++ b/data/os/AlmaLinux/8.yaml
@@ -1,0 +1,39 @@
+---
+ssh::packages:
+  - 'openssh-clients'
+ssh::server::packages:
+  - 'openssh-server'
+
+ssh::include: '/etc/ssh/ssh_config.d/*.conf'
+
+ssh::server::accept_env:
+  - 'LANG'
+  - 'LC_CTYPE'
+  - 'LC_NUMERIC'
+  - 'LC_TIME'
+  - 'LC_COLLATE'
+  - 'LC_MONETARY'
+  - 'LC_MESSAGES'
+  - 'LC_PAPER'
+  - 'LC_NAME'
+  - 'LC_ADDRESS'
+  - 'LC_TELEPHONE'
+  - 'LC_MEASUREMENT'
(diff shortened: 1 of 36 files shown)


anders-larsson:
Hi @Phil-Friderici,

Would you mind adding RedHat 9 as well? It is currently not supported.

Looking through your changes for EL9 distributions they seem to be missing configuration/a decision needs to be made how the .d directories should be handled. Right now they are included by the main configuration file but not maintained. Is this how we want it to work?

Until this is officially supported by the module should all configuration be moved into the main configuration file ignoring the contents of the .d directories?

Phil-Friderici:
> Would you mind adding RedHat 9 as well? It is currently not supported.

Will do so.

> Looking through your changes for EL9 distributions they seem to be missing configuration/a decision needs to be made how the .d directories should be handled. Right now they are included by the main configuration file but not maintained. Is this how we want it to work?

I have started with adding just basic support for the new OS flavours. This should enable users to fully configure SSH in the old style (all configuration inside ssh_config/sshd_config).
You are rights, to support the new style (using include and .d directories) the module would need maintaining these as well.
If wanted I can start working on that later as well. Maybe @ghoneycutt or @treydock can drop there opinion here too ?

> Until this is officially supported by the module should all configuration be moved into the main configuration file ignoring the contents of the .d directories?

For the default configuration I try to touch as less as possible and change nothing from the OS vendor indented settings.
As long as the module doesn't manage .d directory configuration files itself you have to do so by yourself or put all your setings into the old configuration files and remove the include directives.

anders-larsson:
Sounds good to me. It can always be replaced locally by the user. I'll update the Ubuntu 22.04 to support support v4. Our stance than is to maintain the defaults (not setting stuff explicitly)? In previous versions stuff seem to be set explicitly (RedHat 8 for example) instead of using the implicit defaults.

Phil-Friderici:
closed in favour of https://github.com/ghoneycutt/puppet-module-ssh/pull/398

PR #405, opened 2023-04-12
Fix for multiple HostCertificate files

Phil-Friderici:
Multiple files will create one HostCertificate line for each file.

Fixes #391


jfroche:
You might want to run `rake strings:generate:reference` as well to update the REFERENCE.md file using puppet strings

ghoneycutt (approved):

Issue #408, opened 2023-06-12
Move Include to top of sshd_config.erb


enkidushane:
The current alphabetical placement of the Include option in the sshd_config template can lead to confusing results due to SSHs behavior of using "the first obtained value".  Single value options occurring alphabetically before "Include" will be ignored if set differently in an included file, while single value options occurring alphabetically after "Include" will take effect.

In recent versions of Ubuntu, the "Include /etc/ssh/sshd_config.d/*.conf" option is added by default at the very start of the sshd_config file.  Given SSH's "first obtained value" behavior, this placement seems logical,  and moving the "Include" stanza to the top of the sshd_config.erb template would avoid potentially confusing results.

PR #409, opened 2023-06-14
Move Include to top of sshd_config template

enkidushane:
Moves Include option to top of sshd_config template

commit for [#408](https://github.com/ghoneycutt/puppet-module-ssh/issues/408)


ghoneycutt:
This functionality has been added. Thank you!

PR #411, opened 2023-08-08
(docs) Fix ChallengeResponseAuthentication Option

jplindquist:
Fix option name to match template https://github.com/ghoneycutt/puppet-module-ssh/blob/v4.1.0/templates/sshd_config.erb#L56-L58


ghoneycutt (approved):

ghoneycutt:
Thank you!

Issue #407, opened 2023-05-30
notify Service['sshd_service'] when file resource in config_file_server class changes


absld:
Hi,

in class `ssh::server`, there is a subscribe from `service {'sshd_service': }` resource declaration to `File['sshd_config']` resource, which causes the `sshd` service to be restarted after changes in `sshd_config` file.

Instead of defining our `sshd` settings within `sshd_config` file, we use the `Include` mechanism in combination with defining our custom `sshd` settings via `ssh::server::config_files` hash. This results in the configuration parameters get written into `/etc/ssh/sshd_config.d/somefile`. So far so good. 

Unfortunately, there is no notify in the `file` resource within `ssh::config_file_server` class, which is declared in `ssh::server` class for each `$config_files` hash key. So, if configuration parameters change in `/etc/ssh/sshd_config.d/somefile`, the `sshd` service is not refreshed or restarted

PR #413, opened 2023-09-01
Notify service if managed and config_file_server changes

bwitt:
Fixes https://github.com/ghoneycutt/puppet-module-ssh/issues/407


bwitt:
not sure if this will work

PR #415, opened 2023-10-21
Increase stdlib to v9. Increase concat to v9. Update PDK to v3

canihavethisone:
Unit tests pass with Puppet 7.25.  There is a fail with Puppet 8.2.0 so left metadata at v7.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,11 +24,11 @@ jobs:
             fixtures: .fixtures-latest.yml
             allow_failure: true
           - ruby: 3.2.2
-            puppet: 8
+            puppet: 7
             fixtures: .fixtures.yml
             allow_failure: false
           - ruby: 3.2.2
-            puppet: 8
+            puppet: 7
             fixtures: .fixtures-latest.yml
             allow_failure: true
     env:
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yaml
```diff
@@ -15,27 +15,35 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - ruby: 2.7.6
+          - ruby: 2.7.8
```
ghoneycutt, line 18: Why the version change here?

.github/workflows/ci.yaml
```diff
@@ -15,27 +15,35 @@ jobs:
...
+            puppet: 7
+            fixtures: .fixtures-latest.yml
+            allow_failure: true
+          - ruby: 3.2.2
+            puppet: 7
+            fixtures: .fixtures.yml
+            allow_failure: false
+          - ruby: 3.2.2
```
ghoneycutt, line 30: Is puppet 7 running ruby 3.2.2?

spec/acceptance/nodesets/el8.yml
```diff
@@ -1,22 +1,25 @@
 HOSTS:
-  rocky-8:
+  el8:
```
ghoneycutt, line 2: Why the move from rocky to el?

ghoneycutt:
This would be easier to understand if you separated stdlib 9, adding platforms and removing platforms into different PR's.

canihavethisone:
Hey, they were changes made by pdk.  You can either correct them as required or i can replace the pr with one that doesn't upgrade pdk

Issue #427, opened 2024-09-27
sshd Ubuntu 24.04


lucarothmann:
Hello ghoneycutt,

will this module also be available for Ubunt 24.04? 
If so, is there already a time frame for the release?

Best regards 🙂

flichtenheld:
Yeah, I stumbled over this this week as well. Using the defaults for Ubuntu 22.04 seems to work fine, so created PR with that.

PR #428, opened 2024-09-27
Add support for Debian 12 and Ubuntu 24.04

flichtenheld:
Simple copy of Debian 11 and Ubuntu 22.04 config. Saw no reason to change any of the defaults. Definitely better than the global defaults...

Fixes: #427

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -58,8 +58,10 @@ jobs:
           - "el8"
           - "el9"
           - "debian-11"
+          - "debian-12"
           - "ubuntu-2004"
           - "ubuntu-2204"
+          - "ubuntu-2404"
         puppet:
           - "puppet7"
           - "puppet8"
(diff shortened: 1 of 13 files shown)


ghoneycutt:
This data is valid, so people can copy it and use it today before it is merged as this module is forward compatible in that way.

ghoneycutt:
@flichtenheld could you please add the acceptance testing to support this?

flichtenheld:
Will do

flichtenheld:
> @flichtenheld could you please add the acceptance testing to support this?

@ghoneycutt  Added enough to make `pdk test unit` pass. Let me know if there is anything else I should run.

flichtenheld:
Now also enabled GHA tests and rebased on newest main.

ghoneycutt:
Approved the tests and they are coming up with errors.

ghoneycutt:
Thanks @baurmatt and @flichtenheld !

PR #433, opened 2025-01-31
Adding UseRoaming parameter

JasonSorbello:
It looks like the UseRoaming parameter was removed in the v4 rewrite, commit a55f64c, and not added back in. Sorry if this was done on purpose and I'm missing context. There could be other parameters in the same boat too but this is the only one I noticed.

### DIFF
--- a/templates/ssh_config.erb
+++ b/templates/ssh_config.erb
@@ -281,6 +281,9 @@ Host <%= @host %>
 <% if @update_host_keys != nil -%>
   UpdateHostKeys <%= @update_host_keys %>
 <% end -%>
+<% if @use_roaming != nil -%>
+  UseRoaming <%= @use_roaming %>
+<% end -%>
 <% if @user != nil -%>
   User <%= @user %>
 <% end -%>


templates/ssh_config.erb
```diff
@@ -281,6 +281,9 @@ Host <%= @host %>
 <% if @update_host_keys != nil -%>
   UpdateHostKeys <%= @update_host_keys %>
 <% end -%>
+<% if @use_roaming != nil -%>
```
bwitt, line 284: should this go after `user_known_hosts_file` to keep in the same order as https://github.com/ghoneycutt/puppet-module-ssh/blob/main/REFERENCE.md ?

JasonSorbello: Either way is fine. I tried to order it alphabetically but I don't know which comes first.

That said, ssh_config.erb is mostly alphabetical except the first and last entries but REFERENCE.md is seemingly random. Should we make them both alphabetical (except the first and last) so they both match?

ghoneycutt:
This option is no longer listed in the man page (https://man.openbsd.org/ssh_config). You can use it though through the `$custom` parameter.

PR #434, opened 2025-01-31
add X11UseLocalhost parameter to ssh class

tq-t:
Added support for the X11UseLocalhost parameter in the ssh class. Updated ssh_config.erb to include the new parameter when set.


tq-t:
close as  is under ssh::server::x11_use_localhost
https://github.com/ghoneycutt/puppet-module-ssh/blob/main/manifests/server.pp#L458

PR #435, opened 2025-04-09
Remove 18.04 from README

arjenz:
As 18.04 was dropped in https://github.com/ghoneycutt/puppet-module-ssh/pull/417


ghoneycutt:
Thank you!

PR #440, opened 2025-08-31
Add Debian 13 support by reorganizing data for Debian / Ubuntu

Arakmar:
Debian and Ubuntu data has been reorganized to use a per os global data file instead of duplicating it for each new versions.
Apart Ubuntu 20.04 (which is EOL), all settings are the same for each Ubuntu or Debian versions.
Version specific defaults can still be used if needed by a future version.

It also indirectly adds support for Debian 13 which is using the same defaults.


ghoneycutt:
Thank you! Closing in favor of https://github.com/ghoneycutt/puppet-module-ssh/pull/458 where Debian 13 is added.

PR #441, opened 2025-09-02
Add support for Debian 13

mika:
Debian v13 AKA trixie was released on 2025-08-09 as Debian's new stable release.

The default `/etc/ssh/ssh_config` + `/etc/ssh/sshd_config` on Debian 13 are similar to what's present on Debian 12, only `AcceptEnv` now also includes `COLORTERM` + `NO_COLOR`.

FTR: I had the main changes laying around since some weeks for testing internally, only now found time to provide it as PR. Also only now noticed that @Arakmar came up with https://github.com/ghoneycutt/puppet-module-ssh/pull/440/ already just two days ago. :) Not sure which approach is better and easier to maintain (given that different Debian releases ship different defaults, do we want to care?), feel free to drop or take from my PR whatever is needed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,6 +59,7 @@ jobs:
           - "el9"
           - "debian-11"
           - "debian-12"
+          - "debian-13"
           - "ubuntu-2004"
           - "ubuntu-2204"
           - "ubuntu-2404"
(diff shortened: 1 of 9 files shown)


ghoneycutt:
Thank you for this contribution. Closing in favor of https://github.com/ghoneycutt/puppet-module-ssh/pull/458 where I rebased this.

PR #443, opened 2026-01-03
Add RHEL 10 Support

KeithWard:
Mostly the same as EL9

Added a note mentioning that UsePAM: no, is not supported any longer it emits a warning saying it causes problems.


sharumpe:
Came here to look for this same thing.  Would be awesome to get this merged.

ghoneycutt:
Thank you! Superseded by https://github.com/ghoneycutt/puppet-module-ssh/pull/450 which has the testing fixtures.

Issue #437, opened 2025-05-14
ChallengeResponseAuthentication cannot be disabled on Solaris 11


jim-mose:
The current hieradata for Solaris 11 uses the ssh::server::custom parameter to set ChallengeResponseAuthentication to yes. Attempting to override it by setting ssh::server::challenge_response_authentication to "no" causes two ChallengeResponseAuthentication configuration lines to appear in /etc/ssh/sshd_config.

Recommend controlling this setting to use using the ssh::server::challenge_response_authentication parameter within data/os/Solaris/11.yaml, so that it can be overridden elsewhere in users' hiera data.

PR #446, opened 2026-06-01
(ISSUE-437) ChallengeResponseAuthentication cannot be reset

ghoneycutt:
ChallengeResponseAuthentication is being set with the 'custom' parameter. Without this patch, one cannot set this to 'no'.


ghoneycutt:
Fixes https://github.com/ghoneycutt/puppet-module-ssh/issues/437

PR #454, opened 2026-09-04
Add support for Ubuntu 26.04

justmedude:
## Description

Adds Ubuntu 26.04 support, following the same pattern as the existing 22.04/24.04 additions (#427 / the Debian 12 + Ubuntu 24.04 commits): a copy of the 24.04 defaults, since nothing needs to change.

Verified this is actually true rather than just copying blindly:

- Installed `openssh-client`/`openssh-server` on a real Ubuntu 26.04 container and confirmed the package names, the `ssh`/`sshd` service unit names, and the `/usr/lib/openssh/sftp-server` subsystem path are all unchanged from 24.04.
- Mounted the module in a real Puppet environment, applied `include ssh` / `include ssh::server` against that container for real (not just a catalog compile), and diffed the actual rendered `/etc/ssh/ssh_config` and `/etc/ssh/sshd_config` byte-for-byte against the new test fixtures added here -- exact match.
- Confirmed idempotency across three consecutive applies.

## Checklist

- [x] data/os/Ubuntu/26.04.yaml added (copy of 24.04)
- [x] metadata.json updated
- [x] README.md updated
- [x] CI matrix (.github/workflows/ci.yaml) updated
- [x] spec/classes/init_spec.rb and server_spec.rb OS-matching regexes extended
- [x] spec/fixtures/testing/Ubuntu-26.04_{ssh,sshd}_config added
- [x] spec/acceptance/nodesets/ubuntu-2604.yml added


justmedude:
Looked into the CI failures on this PR. Three categories, only one of which is really "new" behavior from adding 26.04 itself -- the other two trace to pinned test-tooling gems not yet knowing about this release:

## Unit tests (`Puppet 7`/`Puppet 8`, both fixture variants) -- `facterdb` has no Ubuntu 26.04 fact data

`Gemfile.lock` pins `facterdb 1.27.0` (capped below 2.0 transitively). That version has no fact fixtures for Ubuntu 26.04 at all, and querying for a release it doesn't have doesn't fail gracefully -- it throws `NoMethodError: undefined method 'map' for nil:NilClass` ("Invalid JSON given"). That breaks every spec file that calls `on_supported_os` without filtering (`init_spec.rb`, `server_spec.rb`, `config_entry_spec.rb`, `config_file_client_spec.rb`, `config_file_server_spec.rb` -- 5 files, matching the "5 errors" in the job output). Reproduced the same error locally outside CI too. Latest `facterdb` is 4.5.0; a real fix means bumping whatever's capping the version below 2.0, which felt like a bigger call than this PR should make unilaterally.

## `acceptance (ubuntu-2604, *)` -- `beaker` hardcodes a package that no longer exists on 26.04

Traced the actual Docker build failure to `E: Package 'ntpdate' has no installation candidate` (Ubuntu 26.04 replaced it with `ntpsec-ntpdate`). The `ntpdate` install comes from `beaker` 4.42.0's `lib/beaker/host_prebuilt_steps.rb`, which hardcodes `UNIX_PACKAGES = ['curl', 'ntpdate']` as the fallback for any platform its `case host['platform']` switch doesn't explicitly handle -- Ubuntu isn't special-cased there, only Debian is, so it falls through to the generic Unix list. There's no nodeset-level override for this anywhere in the gem, so it's not something the `ubuntu-2604.yml` nodeset file added here can work around. This would hit any module's Ubuntu 26.04 acceptance job on this `beaker` version, not just this one.

## `acceptance (debian-11, *)` -- unrelated, pre-existing flakiness

`404 Not Found` fetching a Debian security-mirror package (`libc-l10n`), nothing to do with this change. `main`'s own CI history shows this job intermittently failing already, independent of this PR.

Happy to help however's most useful here -- e.g. splitting the `data/os/Ubuntu/26.04.yaml` + `metadata.json` support out from the CI-matrix/test additions if you'd rather land the actual module support first and handle the tooling bumps separately, or looking into the `facterdb`/`beaker` version bumps as follow-up PRs. Let me know which you'd prefer.

ghoneycutt:
Rebased in https://github.com/ghoneycutt/puppet-module-ssh/pull/457

PR #455, opened 2026-09-07
Modernize testing toolchain (voxpupuli-test 9.2.1, beaker ~> 6.9)

justmedude:
## Summary

Modernizes the pinned `voxpupuli-test` (6.0.0 -> 9.2.1) and `beaker` (~> 4.29 -> ~> 6.9) gems, picking up current `facterdb` fact data and beaker's `Beaker::Platform`-based package resolution, while staying compatible with the Ruby 2.7.8 CI legs this module still tests against.

Opened as a draft to get early feedback / CI signal before final review, and kept separate from other in-flight PRs since this is a general tooling/CI concern rather than an OS-support addition.

## Changes

- **`Gemfile`**: `voxpupuli-test` capped at `9.2.1` and `beaker` at `~> 6.9`, rather than jumping to their absolute latest. Both gems' newest versions require Ruby >= 3.2 transitively (via `puppetlabs_spec_helper >= 9.0.0` and `beaker >= 7.0` respectively), which breaks every CI job still running on Ruby 2.7.8 (the legacy `puppet7` unit leg and all acceptance jobs). These caps are the newest versions that stay Ruby-2.7-compatible: `beaker 6.9.0` already includes the `Beaker::Platform`-based package resolution (with an explicit Ubuntu 26.04/"resolute" mapping) that motivated the bump, and `voxpupuli-test 9.2.1` still resolves `facterdb` to `3.10.0`, fixing the original RedHat 7 fact-data gap.
- **`Gemfile`**: also pin `json` to `~> 2.3.0` on Ruby 2.7 specifically. The bumped `rubocop` (1.50.2, pulled in transitively) depends on a real `json` gem, so Bundler installs `json 2.21.2` instead of relying on Ruby 2.7.8's bundled default (`2.3.0`) -- and that newer `json` breaks Puppet 7.34's deprecated PSON compatibility shim (`NameError: constant PSON::Parser not defined` during `rake syntax`). Confirmed this is new (doesn't happen with the original pins) and that pinning `json` back down fixes it, with no effect on Puppet 8 / Ruby 3.2 where `json` still resolves freely.
- **`Rakefile`**: drop the redundant `require 'puppet-syntax/tasks/puppet-syntax'` — `puppetlabs_spec_helper` has auto-required it itself since at least 8.0.0, so the explicit require was always redundant and caused duplicate `syntax:manifests` task registration.
- **`spec/classes/init_spec.rb`, `spec/defines/config_entry_spec.rb`**: the "OS independent, only test one" fixtures move off RedHat 7 (no fact data even at `facterdb 3.10.0`) to RedHat 8, matching the existing convention already used in `spec/classes/server_spec.rb`.
- **`spec/fixtures/testing/Archlinux-_ssh(d)_config`**: renamed to `Archlinux-6_ssh(d)_config` — `rspec-puppet-facts` resolves Archlinux's release to `"6"` independent of the facterdb version pinned here.

`Gemfile.lock` is gitignored in this repo, so it's not part of this diff; only the `Gemfile` version constraints are changed. `REFERENCE.md` is untouched, since it's only meant to be regenerated by the `:release` rake task.

## Test plan

Verified locally under **both** Ruby 3.2.9 and Ruby 2.7.8 (built specifically to reproduce the CI legs that were failing):

- [x] `bundle exec rake parallel_spec` — 6269 examples, 0 failures, 100% resource coverage (both Ruby versions)
- [x] `bundle exec rake syntax lint validate rubocop metadata_lint` — all clean (both Ruby versions)
- [x] `bundle exec rake beaker` against `ubuntu-2204` under Ruby 2.7.8 — 4 examples, 0 failures, confirming the acceptance path works end-to-end on the Ruby version that was previously failing outright

## Known unrelated CI flake

The `debian-11` acceptance jobs may fail with a 404 fetching `libc-l10n_*.deb` from `deb.debian.org`'s security archive. That's an external Debian 11 (oldstable) mirror availability issue, unrelated to this PR -- same category as Ubuntu 26.04's unpublished `puppet8-release-resolute.deb` blocker on the other open PR.


spec/classes/init_spec.rb
```diff
@@ -379,6 +379,10 @@
 
       it { is_expected.to contain_package('openssh-clients') }
       it { is_expected.to contain_file('ssh_config_include_dir').with_require(['Package[openssh-clients]']) }
+      # only needed for 100% resource coverage: redhat-9's module-default config_files entry
```
ghoneycutt: There is already a specific context for this under the following. Adding it here does not make sense because the context is about the `manage_packages` parameter. If you are just trying to get 100% coverage and not without this, it would be better to just reduce coverage in `spec/spec_helper.rb` than to make the tests harder to understand.

```ruby
context "on #{os} with config_files set to a valid hash" do
```

REFERENCE.md
```diff
@@ -612,7 +612,7 @@ Default value: `undef`
 
 ##### <a name="-ssh--forward_x11_timeout"></a>`forward_x11_timeout`
 
-Data type: `Variant[Undef, String[1], Integer[0]]`
```
ghoneycutt, line 615: Please revert changes to this file as it is auto generated.

spec/classes/init_spec.rb
justmedude: Good catch, thanks. Rather than shoehorning a coverage-only assertion into the `manage_packages` context, I moved the "OS independent" test fixture off RedHat 9 to RedHat 8 (matching the convention already used in `server_spec.rb`), which avoids the config_files-triggered coverage gap entirely — no extra assertion needed. Pushed in 6f2e6af.

REFERENCE.md
justmedude: Reverted, thanks — pushed in 6f2e6af.

ghoneycutt:
This was really helpful and the changes have been incorporated in https://github.com/ghoneycutt/puppet-module-ssh/pull/456

